PR #23, opened 2017-03-17
biomass check [fix #2]

### LINKED ISSUE #2: Test that the sum of molecular masses of biomass precursors adds up to 1 g DCW

ChristianLieven:
Add checks and tests that assert the sum of biomass components to be 1 within an absolute deviation of 1e-03.

### DIFF
--- a/memote/suite/tests/test_biomass.py
+++ b/memote/suite/tests/test_biomass.py
@@ -19,21 +19,23 @@
 
 from __future__ import absolute_import
 
-import logging
+import numpy as np
 
-from memote.support.consistency import (
-    check_stoichiometric_consistency, find_unconserved_metabolites)
+from memote.support.biomass import sum_biomass_weight
 
-LOGGER = logging.getLogger(__name__)
+from memote.support.helpers import find_biomass_reaction
 
 
-def test_stoichiometric_consistency(model, store):
-    """Expect that the metabolic model is mass-balanced."""
-    is_consistent = check_stoichiometric_consistency(model)
-    store["is_consistent"] = is_consistent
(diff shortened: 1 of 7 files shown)


Midnighter:
@ChristianLieven and I were just talking about that yesterday. We clearly need to distinguish them but haven't had a really complete answer yet. Do you have a suggestion for a good approach?

beatrizgj:
I understand a biomass component reaction is a normal reaction, with left and right sides; while the main biomass reaction has only one side. It could be useful to distinguish them.

About the independent biomass tests, at least, 'test_biomass_consistency' I think it is only recommendable for main biomass reaction. Maybe the same for some others.

beatrizgj:
Other option is to remove the individual biomass reactions, simplifying to only the main biomass reaction(s) (one side reaction), replacing the biomass component metabolite in main biomass reaction with left hand side of individual component reactions.
Then, the biomass test would be only required for the wild type (and core, if there is) biomass reaction.

ChristianLieven:
> main biomass reaction has only one side
Note that this is not the case when you consider that the biomass reaction should contain a term for growth-associated maintenance. This will lead to the righthand side of the reaction containing the righthand components of an ATP hydrolysis reaction = 'atp + h2o <-> adp + pi + h'

> replacing the biomass component metabolite in main biomass reaction with left hand side of individual component reactions
This would require us to recalculate the stoichiometry of each of the components according to the overall composition.

I recognise how difficult it is to come up with a solution that accepts all possible forms of implementations.

beatrizgj:
I see, it is difficult to find a solution fits all current models,

PR #98, opened 2017-05-16
fix: require libsbml, update cobra

Midnighter:
`python-libsbml` takes a long time to install but should be a hard dependency
because many models require it for loading. Silent errors are then not very
helpful.


hredestig:
for travis you can
```
pip install python-libsbml==5.12.1 -f https://s3.eu-central-1.amazonaws.com/moonlight-science/wheelhouse/index.html --no-cache-dir
```
to reduce ci time

PR #103, opened 2017-05-23
33 annotation tests

### LINKED ISSUE #33: Test if metabolites can be mapped against MetaNetX DB

ChristianLieven:
resolves #33 

This PR includes a couple of new checks and tests.

First of all when I refer to 'items' I mean either metabolites or reactions.

1. I am adding `annotation.py`, which contains:
`REACTION_ANNOTATIONS` : A dictionary mapping MIRIAM-style reaction database namespaces against the corresponding identifier patterns as compiled regex patterns.
METABOLITE_ANNOTATIONS` : A dictionary mapping MIRIAM-style metabolite database namespaces against the corresponding identifier patterns as compiled regex patterns.

I selected the included databases more or less from experience, i.e. what I had seen to be commonly referenced in metabolic models. I am open for input on what are or aren't 'must-have' databases.

In addition to these two dictionaries, `annotation.py` contains functions to list items without annotation, list items without annotations of each specific database, list items that do contain annotations of a specific database, but which identifiers do not match the corresponding regex pattern.

2. `test_annotation.py`:
Contains annotation specific tests for a given model.
We expect that no item is without annotation, that all items have annotations from all databases and that the identifiers of all of those databases are correct.

3. `test_for_annotation.py`:
Contains test for the code in `annotation.py`.
We expect that no item is without annotation, that all items have annotations from all databases and that the identifiers of all of those databases are correct.

### DIFF
--- a/memote/suite/tests/test_annotation.py
+++ b/memote/suite/tests/test_annotation.py
@@ -131,3 +131,28 @@ def test_rxn_wrong_annotation_ids(read_only_model, store):
             "{}".format(
                 db_id, ", ".join(store['rxn_wrong_annotation_ids'][db_id])
             )
+
+
+def test_met_id_namespace_consistency(read_only_model, store):
+    """
+    Expect metabolite IDs to be from the same namespace.
+    """
+    met_id_namespace = annotation.collect_met_id_namespace(read_only_model)
+    distribution = met_id_namespace[met_id_namespace == 1].count()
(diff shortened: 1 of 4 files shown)


ChristianLieven:
I'll still be updating this by a couple of relevant features. So maybe wait just a little longer with pulling this in.

ChristianLieven:
This is good to go now!

PR #118, opened 2017-06-29
fix: repair `test_biomass` parametrized test functions

Midnighter:
The parametrized `test_biomass` functions were broken because fixtures cannot be used directly inside of the `pytest.mark.parametrize` decorator. Instead we now inject those values into the pytest namespace such that they are available as globals. A bit ugly but seems most feasible.

### DIFF
--- a/memote/suite/collect.py
+++ b/memote/suite/collect.py
@@ -39,6 +39,7 @@ with warnings.catch_warnings():
     from cobra.io import read_sbml_model
 
 from memote.suite.reporting.reports import BasicReport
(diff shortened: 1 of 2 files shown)


ChristianLieven:
That sounds very hacky, is that a consequence of refactoring the model creator function?

Midnighter:
The problem is that `pytest.mark.parametrize` expects a list of values. However, the model itself is provided by a fixture (which makes sense everywhere else) that cannot be used in a non-pytest function or global variable which we could the use in the parametrize decorator. :S Not super pretty but also not terrible.

PR #129, opened 2017-07-20
Dynamic jinja templates

ChristianLieven:
In this PR, I reworked the jinja2 templates for the one-time report to accept all test results without the need to predefine them, filling the AngularMaterial cards dynamically. I used Jquery to build responsive tables that enable the user to toggle further information on given test results. I also swapped the dark 'memote' logo for a white one, and adapted the blue to the colour I've been using in all the figures and the documentation.

Here is a quick look at the finished result:

screen shot 2017-07-20 at 23 36 21

The work here is supposed to be the prototype foundation for a more extensive reworking of how the tests are reported, so that ideally users later only have to worry about writing tests and not about integrating them into the report.


ChristianLieven:
Might be worth to test the jinja2 template logic in the future.

https://stackoverflow.com/questions/42090084/how-can-i-unit-test-jinja2-template-logic

PR #165, opened 2017-08-16
feat: add a replay option

Midnighter:
`memote new` now acccepts a `--replay`` option that will create a repository from the template with the exact answers as before. Should make playing with this feature a little more convenient.


Midnighter:
I merged this because the only failing check was the link check on the docs and BiGG is currently down (so the check cannot pass until BiGG is up again).

Issue #168, opened 2017-08-18
Fixture non_cytosolic may raise a ValueError


Midnighter:
#### Problem description

Simply running the following command in the [memote-demo](https://github.com/ChristianLieven/memote-demo/tree/a4721caf866e29fb079ed1cb8354c23198eed691) repo which contains a model without compartments, raises a `ValueError`. This should be avoided and an empty list returned instead.

#### Code Sample

Code:

```
memote -a "--tb auto"
```

Error:

```bash
@pytest.fixture(scope="module")
    def non_cytosolic(read_only_model, store):
        """Provide all non-cytosolic compartments."""
        compartments = sorted(read_only_model.compartments)
>       compartments.remove('c')
E       ValueError: list.remove(x): x not in list
```

### Context

<details>

System Information
==================
OS                          Linux
OS-release       4.4.0-91-generic
Python                      3.5.2

Package Versions
================
pip                         9.0.1
setuptools                 36.2.7
cobra                       0.8.2
future                     0.16.0
swiglpk                     1.4.4
optlang                     1.2.1
ruamel.yaml               0.14.12
pandas                     0.20.3
numpy                      1.13.1
python-libsbml             5.15.0
click                         6.7
click-configfile            0.2.2
colorama                    0.3.9
pytest                      3.2.1
dask                       0.15.1
cloudpickle                 0.4.0
toolz                       0.8.2
Jinja2                      2.9.6
jinja2-ospath               0.2.1
cookiecutter                1.5.1
plotly                     2.0.14

</details>

PR #169, opened 2017-08-18
fix: catch errors where they have been reported

### LINKED ISSUE #145: Catch potential errors in find_stoichiometrically_balanced_cycles!
find_stoichiometrically_balanced_cycles errors which leads to the test results being omitted from the report. 

It should be clear to the user why the test errors.

ChristianLieven:
* [x] fix #145, #168
* [x] This quick fix should catch the value error raises when handing memote a model without any compartments or without the compartment 'c' for cytosol. Moreover, this fix now catches errors that might occur when running FVA and loopless FVA, which prohibited the corresponding test to show up.
* [x] All tests passed locally
* [x] Update History.rst

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,10 +4,8 @@ History
 Next Release
 ------------
 
-0.4.3 (2017-08-26)
-------------------
-
 * Temporarily remove ``test_find_stoichiometrically_balanced_cycles``
+* Catch errors when testing for compartments and loops.
 
 0.4.2 (2017-08-22)
 ------------------
--- a/memote/suite/tests/test_syntax.py
+++ b/memote/suite/tests/test_syntax.py
@@ -32,10 +32,10 @@ def non_cytosolic(read_only_model, store):
         store["non_cytosolic"] = compartments.remove('c')
         return compartments
     except ValueError:
-        print ("The model does not contain a compartment ID "
-               "labeled ``c`` for the cytosol which is an essential"
-               "compartment. Many syntax tests depend on this being labeled"
-               "accordingly.")
+        print("The model does not contain a compartment ID "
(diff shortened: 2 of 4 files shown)


ChristianLieven:
Something isn't right with this branch! Do not pull it in yet. When I run `memote report snapshot <model>` with this branch now it seems to just crash when it gets to `test_find_stoichiometrically_balanced_cycles`. Not sure what is causing this exactly because all the pytests for the code pass with no problems.

Midnighter:
Well, a complete FVA and a loopless FVA for a large model is a challenge. Yin Hoon tried this for a Recon model and it also failed. So let's look into that test further.

ChristianLieven:
The complete FVA on iJO1366 runs in less than 10 seconds, however the loopless FVA completely stops dead. I've now tried it in a notebook and just on python 2 in the terminal but it doesn't work either way. This is with Version: 0.8.2 of Cobrapy. Could you give this a try and tell me if it also takes infinitely long for you:

`from cobra.io import read_sbml_model
from cobra.flux_analysis import flux_variability_analysis
model = read_sbml_model("/Users/clie/Desktop/Side_Projects/Memote/Ecoli_Model_collection/iJO1366/iJO1366.xml")`
fva_result_loopless = flux_variability_analysis(model, loopless=True)

Midnighter:
Finished for me after a couple of minutes.

ChristianLieven:
Phew. Well at least it finished, but yeah that's still way to long imo for a quick snapshot test. I'm gonna open a new PR to temporarily comment out that test, while I try to optimize the runtime/ find loops in an alternative way.

Midnighter:
Again merged even though `docs` failed. Hope the pathway tools issue will be resolved after the weekend.

PR #179, opened 2017-08-23
Feat/collect test status

### LINKED ISSUE #170: Collect broad performance for the report (failed or passed)
#### Problem description

Right now the reports are very technical showing the precise results of the tests but not whether the overall test passed or failed. In order to achieve that we'll have to start collecting this information from pytest.

ChristianLieven:
* [x] fix #170
* [x] In this PR, I focused on implementing a way to obtain and store the pytest metrics of a test's performance. I am accessing these metrics through 2 hooks `pytest_runtest_protocol()` and `pytest_report_teststatus()`. With the former, I collect the test function docstring selecting only the summary section, while with the latter I collect the time it took to run the test and the final outcome [passed, failed, skipped]. The results are stored in the `_data` attribute of the `ResultsCollectionPlugin()` object which is instanced when running the `test_model()` function in `api.py`. The structure of the `_data` attribute now looks like this:
`self._data["report"][">test_module<"][">test_function<"] = {
    "description": docstrings,
    "result": pass/skip/fail,
    "data": extra information that we pass in
}`
`>test_module<` represents the module ID string that the test function code is in such as "test_basic" from `test_basic.py`.
`>test_function<` subsequently represents the function ID such as "test_genes_presence" within `test_basic.py`.

The output for the snapshot report has been adapted to handle these changes, the history report however hasn't been adapted and will likely error. Before rewriting it to handle the new way the data is stored internally, I'd prefer to first establish our desired changes in the internal result/data management and display architecture (nosqlite + angularJS).

I'll keep on adding to this PR to make the last required changes (testing and history.rst)


memote/suite/collect.py
```diff
@@ -109,6 +110,56 @@ def _collect_git_info(self):
...
+        item_name = str(report.location[-1]).lower()
+
+        if module_name not in self._data:
+            self._data[module_name] = dict()
+
+        if item_name not in self._data[module_name]:
+            self._data[module_name][item_name] = dict()
+
+        self._data[module_name][item_name]["duration"] = report.duration
+        self._data[module_name][item_name]["result"] = report.outcome
```
Midnighter: I think you have to be careful with the report outcome. If it was an error then simply returning it will raise the error at this point (if I remember corrrectly). Please make sure this isn't the case or handle appropriately.

Midnighter:
One general comment: We don't currently use them but if we ever utilize test classes then the two hooks need to be changed to account for that. I'm not saying you should do that now. We need to keep that in mind, though.

memote/suite/collect.py
ChristianLieven: Good point! I'm writing tests for this PR as we speak. I'll try to cover that there and adapt my code if necessary.

Midnighter:
Also relevant is that we should start testing the plugin. https://docs.pytest.org/en/latest/writing_plugins.html#testing-plugins

ChristianLieven:
That's very relevant indeed. Since all the added code are directly related to the plugin itself this is probably better than going the route via the api, which is what I had initially planned to do for the tests here.

memote/suite/collect.py
Midnighter: Seems not to be the case https://docs.pytest.org/en/latest/writing_plugins.html#_pytest.runner.TestReport.outcome but when failed it'd be nice to get the error message as well.

Midnighter: I'm also pretty sure that the `pytest_runtest_protocol` runs first and will set up the dictionary structure. It should therefore not be necessary to set it up again in this hook. (Although it doesn't hurt to be careful.)

ChristianLieven: While testing this when I wrote it, not setting the dictionary in each hook would error and I couldn't quite figure out why. So this was the only way it worked for me.

Also, getting the error message should be as simple as also storing this `report.longrepr` as `longrepr = None¶
None or a failure representation.`

memote/suite/collect.py
```diff
@@ -109,6 +110,56 @@ def _collect_git_info(self):
...
+        if item_name not in self._data[module_name]:
+            self._data[module_name][item_name] = dict()
+
+        if item.obj.__doc__:
+            docstring = " ".join(FunctionDoc(item.obj)["Summary"])
+            self._data[module_name][item_name]["summary"] = docstring
+        else:
+            self._data[module_name][item_name]["summary"] = "No summary " \
+                                                            "available."
+
```
Midnighter: This is a perfect use-case for `dict.setdefault` btw. I would rewrite to the following:
```python
module = self._data.setdefault(item.obj.__module__, dict())
case = module.setdefault(str(item.name).lower(), dict())
case["summary"] = docstring
```

ChristianLieven:
Hey Moritz, could you take another look at this. I think I'm mostly done with the results collection, except for the first two tests in Test_Syntax.py where I'm not exactly sure how to access a fixture from within pytest.mark.parametrize.

Cheers!

Midnighter:
We broke Travis. It doesn't like this PR.

PR #180, opened 2017-08-26
refactor: disable test for cycles temporarily

ChristianLieven:
* [X] The test for stoichiometrically balanced cycles takes forever to run, I'm temporarily disabling it to optimise its performance without it further affecting the functionality of memote.
* [X] I commented out the functions `test_find_stoichiometrically_balanced_cycles` in `test_consistency.py` and `test_for_consistency` respectively.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,11 @@ History
 Next Release
 ------------
 
+0.4.3 (2017-08-26)
+------------------
+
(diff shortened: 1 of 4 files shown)


Midnighter:
* I rather marked the offending test as skipped and keep the code in uncommented.
* The pathway tools domain is currently unavailable. I just merged anyway but we should keep an eye on that.

PR #181, opened 2017-08-28
Bug fixes

### LINKED ISSUE #134: Gam in Biomass results are wrong in one-time report
### Description

Isolated functions succeed in identifying the GAM in biomass, yet in the report the output is incorrect when tested for iJO1366.

### LINKED ISSUE #137: logic in test_rxn_ids_in_each_namespace is not able to differentiate between biocyc vs bigg ids
### Description
The regex pattern for biocyc is much broader than the pattern for bigg.reaction, which leads to duplicate matches. With these kind of duplicate matches it is more likely that we have a bigg.reaction ID and we can safely ignore the supposed match to the biocyc namespace.

`
{'brenda': 0, 'ec-code': 0, 'biocyc': 1, 'kegg.reaction': 0, 'rhea': 0, 'bigg.reaction': 1, 'metanetx.reaction': 0}
`

This seems to be working fine with metabolite IDs:

`{'inchikey': 0, 'metanetx.chemical': 0, 'biocyc': 0, 'bigg.metabolite': 11, 'seed.compound': 0, 'kegg.compound': 0, 'hmdb': 0, 'reactome': 0, 'chebi': 0}`

### LINKED ISSUE #140: GTP/GDP driven fail 'abc' syntax check
Reactions such as this don't get recognized correctly by the 'abc' syntax test and instead get flagged as untagged normal transport reactions.

Iron (II) transport via ABC system (GTP) (periplasm)
`fe2_p + gtp_c + h2o_c --> fe2_c + gdp_c + h_c + pi_c`

### LINKED ISSUE #154: Expand set of DB namespaces for annotation tests!
Include Pubchem
Compound would be favoured: http://www.ebi.ac.uk/miriam/main/collections/MIR:00000034

ChristianLieven:
* [x] fix #140, fix #137, fix #134, fix #154 
* [x] With this pull-request I tried to address the bugs and requests we've had in some of the issues:
- The biocyc identifier regex is very general and thus matches any of the other databases. I've implemented a filter for biocyc values in `generate_component_id_namespace_overview` such that whenever an annotation ID matches agains the biocyc regex and any other regex the biocyc match will be removed.
- Previously, memote did not consider abc-transporters to be driven by energy from the hydrolysis of NTPs other than ATP. This was fixed by allowing CTP and GTP as possible driving factors.
- The check for GAM in biomass was made more precise by specifically looking for "atp_c" and "h2o_c" on the reactant side and for "adp_c", "pi_c" and "h_c" on the product side of the bionass reaction. Like with the previous check, I am aware that this favours models with BiGG IDs.
- Expand the required metabolite annotations to include PubChem IDs.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,10 @@ History
 Next Release
 ------------
 
+* Fix a bunch of bugs:
+    - Remove false positive detection of Biocyc annotation
+    - Allow memote to identify CTP or GTP driven transport reactions
+    - Refactor how memote detects GAM in the biomass reaction
 * Add tests to find deadend, orphan and disconnected metabolites.
 * Extend and improve algorithm to find energy-generating cycles
 * Remove the ``print`` statement from ``memote.support.annotation
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -120,7 +120,7 @@ file. Automatic draft reconstruction tools such as `Pathway Tools`_,
 this format. Model repositories such as `BiGG`_ or `BioModels`_ further serve
 as a great resource for models in the correct format.
 
-.. _Pathway Tools: http://bioinformatics.ai.sri.com/ptools/
+.. _Pathway Tools: http://www.pathwaytools.org/
 .. _Model SEED: http://modelseed.org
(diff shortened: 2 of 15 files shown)


ChristianLieven:
Looks like SRI's websites are still down.

Issue #218, opened 2017-09-29
test_find_deadends should consider reaction reversibility


cdanielmachado:
**memote** is reporting that metabolite B in following example is a dead end:

A -> B
C <-> B 

Well... it isn't... :)

PR #224, opened 2017-09-29
fix: account for reversibility

### LINKED ISSUE #217: exchange metabolites considerer orphan metabolites
I am getting a report that all exchange metabolites are orphan metabolites.

The metabolites use BiGG notation (eg: 'M_h2o_e'), and the reactions as well ('R_EX_h2o_e'). Since **memote** is assuming BiGG notation as the default standard it should correctly identify the exchange metabolites and exchange reactions.

Midnighter:
* [x] fix #217, #218
* [x] test for dead-end and orphan metabolites was ignoring reversibility
* [x] added a test model `reversible_gap` to `test_for_consistency`

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,9 @@ History
 Next Release
 ------------
 
+* Correctly account for reversibility when testing for dead-end and orphan
+  metabolites.
+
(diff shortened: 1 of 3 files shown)


ChristianLieven:
Looks good!

PR #251, opened 2017-10-22
refactor: use a decorator instead of model builder

### LINKED ISSUE #248: Create a decorator for registering minimal test models for the test cases
Currently, we have a `model_builder` function that uses a dictionary to retrieve the correct model building function. Maintaining that dictionary is quite tedious. Instead, we should use a decorator on those functions to directly add them to a registry.

Midnighter:
* [x] fix #248

Functions that create models in the `test_for*` pytest modules should now be decorated with `@register_with(MODEL_REGISTRY)` rather than made known in the `model_builder` function.

### DIFF
--- a/memote/utils.py
+++ b/memote/utils.py
@@ -15,27 +15,29 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Ensure the expected functioning of ``memote.suite.cli.reports``."""
+"""Utility functions used by memote and its tests."""
 
 from __future__ import absolute_import
 
-from os.path import exists
-
-from memote.suite.cli.reports import report
+__all__ = ("register_with",)
 
 
-def test_report(runner):
-    """Expect a simple memote report invocation to be successful."""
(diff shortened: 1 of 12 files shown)


memote/utils.py
```diff
@@ -0,0 +1,43 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Utility functions used by memote and its tests."""
+
+from __future__ import absolute_import
+
+__all__ = ("register_with",)
```
ChristianLieven: I've noticed that you add a 'hanging' comma at the end of lists a lot. What is the reason for that?

Midnighter: It's a one-tuple (using `()` rather than `[]`) and it's required for tuples with only one element. Otherwise they are just parentheses around an object.

PR #271, opened 2017-12-01
Angular app

### LINKED ISSUE #265: Create an Angular app for the snapshot report
Done: An Angular app with relevant components and material design is integrated in memote.

ChristianLieven:
* [X] fix #265 (issue number)

I've finished replicating the snapshot report remplate that we've previously with angular 5. The styling of the report is now done with Angular Material2 (https://github.com/angular/material2) and Angular flex-layout (https://github.com/angular/flex-layout).

So the basic scaffold should be good to go. We'll still have to figure out how to minify/compress/bundle this and I guess I'll have to sort out what seems to be like a branching error too before this can be merged.


ChristianLieven:
Fixed the branching error by rebasing. This is now a clean branch starting just at the implementation of the angular report.

memote-report-app/src/app/report-data.service.ts
```diff
@@ -0,0 +1,62 @@
...
+    return this.allTests.find(x => x.id === string);
+  }
+
+  public getString(object){
+    return JSON.stringify(object);
+  }
+
+  private convertResults(data:Object): void{
+    // Store each test result as a TestResult object in a central list.
+    for (let test of Object.keys(data["tests"])){
```
pankaj28843: @ChristianLieven You should use `const test of ...` instead of `let test of ...` because you don't intend to reassign `test inside the for-block.

memote-report-app/src/app/report-data.service.ts
```diff
@@ -0,0 +1,62 @@
...
+        this.statisticsCards.push(
+          new ResultCard(
+            card,
+            data["cards"][card]["title"],
+            data["cards"][card]["cases"]
+          )
+        )
+      }
+    }
+    console.log(this.statisticsCards)
```
pankaj28843: @ChristianLieven You may want to remove this `console.log` statement.

memote-report-app/src/app/test-result.model.ts
```diff
@@ -0,0 +1,24 @@
...
+      this.id = id;
+      this.data = data;
+      this.duration= duration;
+      this.message= message;
+      this.metric= metric;
+      this.result= result;
+      this.summary= summary;
+      this.title= title;
+      this.type= type;
+     }
```
pankaj28843: @ChristianLieven I would rather have the following signature on the constructor method:

```typescript
constructor(id: string, {data, duration, message, metric, result, summary, title, type}: {
    data: any, duration: number, message: string, metric: number,
    result: string, summary: string, title: string, type: string
  }) {
      this.id = id;
      this.data = data;
      this.duration= duration;
      this.message= message;
      this.metric= metric;
      this.result= result;
      this.summary= summary;
      this.title= title;
      this.type= type;
     }
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring

memote-report-app/src/app/report-data.service.ts
```diff
@@ -0,0 +1,62 @@
...
+          test,
+          data["tests"][test]['data'],
+          data["tests"][test]['duration'],
+          data["tests"][test]['message'],
+          data["tests"][test]['metric'],
+          data["tests"][test]['result'],
+          data["tests"][test]['summary'],
+          data["tests"][test]['title'],
+         data["tests"][test]['type']));
+    };
```
pankaj28843: @ChristianLieven Related comment: https://github.com/opencobra/memote/pull/271/files#r154901166

```typescript
new TestResult(test, data["tests"][test])
```

This looks more neat and maintainable to me, you don't need to worry about the order of arguments anymore.

pankaj28843: @ChristianLieven semicolon after a for block is unnecessary, If you configure your Webstorm to use TSLint, it shows warnings like these :smile:

memote-report-app/src/app/report-data.service.ts
```diff
@@ -0,0 +1,62 @@
...
+      if (card === "scored"){
+        this.scoredCard = data["cards"]["scored"]
+      }
+      else{
+        this.statisticsCards.push(
+          new ResultCard(
+            card,
+            data["cards"][card]["title"],
+            data["cards"][card]["cases"]
+          )
```
pankaj28843: @ChristianLieven You can do the same improvement as `TestResult` in this case as well using object-destructuring.

memote-report-app/src/app/report-data.service.ts
ChristianLieven: I don't have Webstorm unfortunately :( But I'll see what I can do about it in Atom!

PR #281, opened 2017-12-19
Implement dedicated SBO test module as outlined in #276

### LINKED ISSUE #276: Tests enforcing the use of SBO terms to annotate model components
#### Problem description
Many coauthors have criticised the initially strong focus of memote on syntax rules. Instead we should benchmark if a model's components are sufficiently and correctly annotated with SBO terms. 

#### Test which should be implemented to complete this issue
- [x] Check if each component for the presence of (any) SBO term annotation
- [x] Specifically check if metabolic reactions are annotated as SBO:0000176
- [x] Specifically check if transport reactions are annotated as SBO:0000185
- [x] Specifically check if metabolites are annotated as SBO:0000247
- [x] Specifically check if genes are annotated as SBO:0000243
- [x] Specifically check if exchange reactions are annotated as SBO: 0000627
- [x] Specifically check if demand reactions are annotated as SBO: 0000628
- [x] Specifically check if sink reactions are annotated as SBO: 0000632
- [x] Specifically check if biomass reactions are annotated as SBO: 0000629

ChristianLieven:
* [x] fix #276
* [x] Adds an array of test for common SBO terms
* [x] all functions are accounted for through unittests

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,7 @@ History
 Next Release
 ------------
 * Let Travis re-package the snapshot report with every release.
+* Add new module to test for the presence of SBO term annotations.
 
 0.5.0 (2018-01-16)
 ------------------
--- a/memote/suite/tests/test_sbo.py
+++ b/memote/suite/tests/test_sbo.py
@@ -1,0 +1,91 @@
+# -*- coding: utf-8 -*-
+
+# Copyright 2017 Novo Nordisk Foundation Center for Biosustainability,
+# Technical University of Denmark.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


ChristianLieven:
Hardcode respective parts of ontology tree for now!

ChristianLieven:
Will continue with this once we find a suitable alternative to owlready (which is not backwards-compatible to python2).

ChristianLieven:
@Midnighter Could we close this PR but keep an issue open? Having some tests in the code is better than no tests and with the speed at which new PRs come in I'd like to avoid this becoming stale. We can still optimise the underlying logic later on.

Midnighter:
Yeah, sounds good. Last I checked ontospy was horribly out-dated. If that's really the case I'd maybe put in the effort to update owlready2 ourselves if necessary. And I think by closing you mean merging?

ChristianLieven:
Yeah merging it. Or if not at least before the other PRs come in. Actually this PR in particular (#316) changes the way the 'type' parameter works, so if that gets merged first the tests for all the other ones will fail. Just fyi!

How do I tell if it is outdated?
The last commit was 17 Nov 2017

Midnighter:
We might get a few complaints that SBO terms are there but not read in properly by cobrapy but here goes :)

PR #289, opened 2017-12-21
chore: update config with full range of tests

ChristianLieven:
This is just a quick one really no major changes. I just made sure to include all the tests we currently offer in the config grouped like we (@Midnighter  and I) discussed previously.


Midnighter:
Are you going to tackle the "miscellaneous" card here as well or in a different PR?

ChristianLieven:
In this PR I've now also addressed something that I previously discussed with @Midnighter:  Tests that are not explicitly assigned to cards/sections of the snapshot report via the configuration file are now automatically collected in a 'Misc. Tests' card. As a consequence of this I addressed a bug which lead to skipped tests not being displayed correctly in the report.

To test these changes I had to overwrite the previous `testData.json` which contributes significantly to the seemingly bloated line changes. In the future, this file could be reduced in size for testing.

Issue #297, opened 2018-01-15
Make memote agnostic of component ID namespaces!


ChristianLieven:
#### Problem description

Memote is quite dependent on the BIGG ID namespace right now. This is unacceptable for many reasons. I'd like to come up with a heuristic for identifying certain components of the model:
- Identify key metabolites #242
- Identify biomass reaction(s) #216
- Identify non-growth associated maintenance reaction(s) #222

Issue #242, opened 2017-10-13
Check for the presence of universal components in the biomass reaction


ChristianLieven:
There are universal components of life that make up the biomass of all organisms known. We should check that these components are contained in the biomass reaction.

Amino Acids, Desoxyribonucleotides, Nucleotides and universal cofactors come to mind.

joanarcx:
Let me know where ou would like me to send the lists.

ChristianLieven:
Hi @joanarcx,

If they are not too large you could post them here? Or if you prefer you can email them to me.

joanarcx:
Sure thing. As per Xavier et al.:

**Aminoacids**:
trp__L,cys__L,his__L,tyr__L,met__L,phe__L,ser__L,pro__L,asp__L,thr__L,gln__L,glu__L,ile__L,arg__L,lys__L,val__L,leu__L,ala__L,gly,asn__L
**DNA**: datp,dctp,dttp,dgtp
**RNA**: ctp, utp, gtp
**Cofactors**: nad, nadp, amet, fad, pydx5p, coa, thmpp, fmn
[I am not including here C1 carriers for it would be a complex test with ORs, although we identified the class as essential, therefore indicating that one should be present - see Fig.3 in the paper if interested.]
**Others:** atp, h2o

joanarcx:
I can email you KEGG and ModelSEED translations for these also, in case you decide to include these namespaces in the testings or even just in the Biomass tests.

Midnighter:
Do you have the metanetx IDs for these?

joanarcx:
no. But it shouldn't be hard to get them.

ChristianLieven:
Yeah, I think thats pretty straightforward! Thank you @joanarcx !!

joanarcx:
Happy to contribute. Yes for MetaNetX you can use the mapping file chem_xref.tsv provided in https://www.metanetx.org/mnxdoc/mnxref.html
it has the direct mappings for BiGG.

kcorreia:
> Cofactors: nad, nadp, amet, fad, pydx5p, coa, thmpp, fmn

Many organisms cannot synthesize all cofactors/vitamins (or other biomass precursors for that matter). I agree most of these should be included in the biomass equation, but consider the penalty you'd get if you have a test for a [low ratio of direct metabolites](https://github.com/opencobra/memote/issues/257).

Some vitamins are also condition-specific (or more precisely some enzymes require vitamins).

Issue #216, opened 2017-09-29
biomass detection


cdanielmachado:
The heuristics for detecting biomass reaction could be a bit more flexible.

A few suggestions:

- more options for name testing ("biomass", "growth", etc)

- combination of objective coefficient and number of substrates (for instance: if ATP production was the objective function, setting the threshold to a minimum of 4-5 substrates would in principle exclude it)

- alternatively, discard the assumption that the model needs to contain a biomass equation, and just make sure that the objective function (whatever it might be) is feasible under suitable conditions

ChristianLieven:
I'll gladly add more options to try and identify the biomass reaction, although changing the ID for the reaction is a super simple fix for model builders. In that sense I'd rather attempt to unify what this particular reaction is called by leaving the heuristic strict. But I'm waiting to hear more opinions on this.

I'm not convinced of the other suggestions:
1. Some lumped reactions might hypothetically consume +5 substrate, consume ATP and be the objective function (in a production strain for instance), hence basing the test on these criteria might not be robust enough.
2. Any cell be it by itself or in a community/ organ should need to construct structural macromolecules to function and by extension a biomass reaction would have to be present. The only exception that comes to my mind here are erythrocytes. But frankly, I cannot asses how common models without a biomass reaction would be. 

So, my suggestion is to detect it by preferably one very robust criterion. I'm happy to rely on the community to discuss and agree on whatever that may be. If this identification for a biomass reaction cannot be identified in a model, all tests pertaining to the Biomass reaction and precursor production are skipped.

Midnighter:
My preferred way would be an annotation with the SBO term. Then we don't have to care what name is assigned at all.

cdanielmachado:
I also think SBO term is a good idea but, it would be nice if the tests still run when the SBO term is missing. 
 
I think memote should be flexible enough to accommodate different modeling practices, rather than enforcing personal preferences. 

Regarding the exact identifiers, it wouldn't be difficult to do a quick survey of common names people have used and include all few more options.

For instance, I prefer to call the reaction "growth" rather than "biomass", since biomass is the product of the process (passive element), not the process itself (active element).

PR #321, opened 2018-01-21
Agnostic heuristic

### LINKED ISSUE #222: Missing NGAM wrongly reported
**memote** is reporting missing NGAM when the lower bound of R_ATPM is zero.

I think it should check for the presence of NGAM but not force a particular value.

The principle of constraint-based modeling is that one iteratively refines a model by applying constraints and reducing the solution space as more data is acquired. The model is valid as long as the "true" phenotype is still somewhere inside the solution space. 

If NGAM is unknown because it was not yet calibrated from experimental data, a range of [0, max_flux] is still a valid constraint.

ChristianLieven:
Fixes #222, #216, #297, #242

Lots of additions, changes and fixes in this one:

- Add a cross-reference shortlist using MetaNetX flatfiles
- Add a script that can be used to add more metabolites and then to re-generate the shortlist
- Add helper function `find_met_in_model` which looks up a query metabolite ID using the MNX namespace in the shortlist and:
    - If no compartment is provided, returns a list of all possible candidates metabolites
    - If a compartment is provided, tries to return a list containing only ONE corresponding metabolite.
- Add helper function `find_compartment_id_in_model` to identify compartments using an internal shortlist of possible compartment names.
- Provide tests for each function
- Refactor code to use these functions specifically:
    - `find_ngam`
    - `find_biomass_reaction`
    - `detect_energy_generating_cycles`
    - `find_exchange_rxns`
    - `find_demand_rxns`
    - `find_sink_rxns`
    - `gam_in_biomass`
    - `find_biomass_precursors`
- Improve `find_ngam` in addition to agnostically looking for ATP hydrolysis reactions, the test now also looks for a range of possible "buzzwords" in the reaction NAME: ['maintenance', 'atpm', 'requirement', 'ngam', 'non-growth', 'associated']. One match suffices as a classification.
- Improve `find_biomass_reaction` to look for three attributes in a biomass reaction, one of which is sufficient to classify it as a biomass reaction:

    1. "Buzzwords" in the reaction ID: ['biomass', 'growth', 'bof']
    2. An annotation matching the SBO-Term SBO:0000630 specifically!
    3. Containing a metabolite matching the regex: `^biomass(_[a-zA-Z]+?)*?$` (case-insensitive)
- Add function `bundle_biomass_components` to identify whether a given biomass reaction is 'split' or 'lumped'. This function looks simply at the size of the biomass reaction. Based on a guess-timated cut-off the reaction is then classified. If it is 'lumped' it is returned without changes, if it is 'split' the reactions of any non-energy precursor metabolite are returned as well. This is based on the assumption that a 'split' biomass reaction has the following structure:
a (1 gDW ash) + b (1 gDW phospholipids) + c (free fatty acids)+ d (1 gDW carbs) + e (1 gDW protein) + f (1 gDW RNA) + g (1 gDW DNA) + h (vitamins/cofactors)-> 1 gDCW biomass. We're supposing that for each macromolecule precursor metabolite there is a single reaction defining its composition i.e. `e` = protein would have the reaction: ` alanine + asparagine + ... + valine --> e`
- Add function, test and model test to identify missing essential precursors to the biomass reaction. The function is `essential_precursors_not_in_biomass`

* [X] fix #x (issue number)
* [X] tests added/passed


memote/support/basic.py
```diff
@@ -104,11 +106,38 @@ def find_ngam(model):
...
+    products = {
+        helpers.find_met_in_model(model, "MNXM7", id_of_main_compartment)[0],
+        helpers.find_met_in_model(model, "MNXM1", id_of_main_compartment)[0],
+        helpers.find_met_in_model(model, "MNXM9", id_of_main_compartment)[0]
+    }
+
+    candidates = [rxn for rxn in atp_adp_conv_rxns
+                  if set(rxn.reactants) == reactants and
+                  set(rxn.products) == products and
+                  rxn.reversibility is False]
```
Midnighter: If you move the last reversibility check to the first position it will help the performance of this test since the set based conditions might unnecessary.

memote/support/biomass.py
```diff
@@ -133,6 +177,143 @@ def find_direct_metabolites(model, reaction):
...
+    e.g. Protein, Phospholipids etc. Each of these have their own composing
+    reactions. Hence we include the reactions of these metabolites in the
+    set that ultimately makes up the returned list of reactions that together
+    make up the biomass equation.
+
+    """
+    if len(reaction.metabolites) >= 16:
+        return [reaction]
+
+    if len(reaction.metabolites) <= 15:
```
Midnighter: This condition is unnecessary. If `len(reaction.metabolites) >= 16` it will never be executed.

memote/support/consistency.py
```diff
@@ -308,30 +312,50 @@ def detect_energy_generating_cycles(model, metabolite_id):
...
+        ENERGY_COUPLES[metabolite_id],
+        main_comp)[0]
 
     dissipation_rxn.add_metabolites(
-        {met.id: -1, dissipation_product: 1})
+        {met: -1, dissipation_product: 1})
+    helpers.close_boundaries_sensibly(model)
+    model.add_reactions([dissipation_rxn])
     model.objective = dissipation_rxn
     solution = model.optimize()
```
Midnighter: You can pass the parameter `model.optimize(raise_error=True)` which will give you an optimization error if the status is not optimal with an appropriate message. You can then use it just like you do with the `RuntimeError` below.

memote/support/helpers.py
```diff
@@ -32,10 +34,48 @@
...
+# different namespaces.
+directory = os.path.dirname(__file__)
+METANETX_SHORTLIST = pd.read_json(
+    directory + "/data/met_id_shortlist.json"
+)
+
+
+# Provide a compartment shortlist to identify specfic compartments whenever
+# necessary.
+COMPARTMENT_SHORTLIST = {
```
Midnighter: Would it also make sense to outsource this to a file?

memote/support/helpers.py
```diff
@@ -32,10 +34,48 @@
...
+import memote.utils as utils
 
 LOGGER = logging.getLogger(__name__)
 
 
+# Read the MetaNetX shortlist to identify specific metabolite IDs across
+# different namespaces.
+directory = os.path.dirname(__file__)
+METANETX_SHORTLIST = pd.read_json(
+    directory + "/data/met_id_shortlist.json"
```
Midnighter: Would YAML be an easier to manually handle format?

memote/support/helpers.py
```diff
@@ -173,8 +208,32 @@ def find_biomass_reaction(model):
...
+
     """
-    return [rxn for rxn in model.reactions if "biomass" in rxn.id.lower()]
+    buzzwords = ['biomass', 'growth', 'bof']
+
+    buzzword_matches = set([rxn for rxn in model.reactions if any(
+        string in rxn.id.lower() for string in buzzwords
+    )])
+
+    sbo_matches = set([rxn for rxn in model.reactions if
```
Midnighter: I would have checked for SBO presence first and only perform the other queries if that didn't yield results.

memote/support/helpers.py
```diff
@@ -173,8 +208,32 @@ def find_biomass_reaction(model):
...
 
+    Returns
+    -------
+    list of biomass reactions
+
     """
-    return [rxn for rxn in model.reactions if "biomass" in rxn.id.lower()]
+    buzzwords = ['biomass', 'growth', 'bof']
+
+    buzzword_matches = set([rxn for rxn in model.reactions if any(
```
Midnighter: It'd be better to also check the name attribute.

memote/support/helpers.py
```diff
@@ -173,8 +208,32 @@ def find_biomass_reaction(model):
...
+        string in rxn.id.lower() for string in buzzwords
+    )])
+
+    sbo_matches = set([rxn for rxn in model.reactions if
+                       rxn.annotation is not None and
+                       'SBO' in rxn.annotation and
+                       rxn.annotation['SBO'] == 'SBO:0000630'])
+
+    regex = re.compile('^{}(_[a-zA-Z]+?)*?$'.format('biomass'), re.IGNORECASE)
```
Midnighter: I would use the same method as you did above for `id` and `name` or query with a simpler regex. Regex's are expensive, though, so overall maybe:
```python
for element in model.reactions|metabolites:
    for word in buzzwords:
        if word in element.id.lower() or word in element.name.lower():
            hits.append(element)
```

memote/support/helpers.py
```diff
@@ -173,8 +208,32 @@ def find_biomass_reaction(model):
...
+    biomass_metabolite = model.metabolites.query(regex)
```
Midnighter: `.query` unless otherwise instructed only checks the `.id` attribute.

memote/support/helpers.py
```diff
@@ -389,16 +451,181 @@ def open_boundaries(model):
...
+    for name in COMPARTMENT_SHORTLIST[compartment_id]:
+        for c_id, c_name in model.compartments.items():
+            if c_name.lower() == name:
+                return c_id
+
+    if compartment_id == 'c':
+        return largest_compartment_id_met(model)
+
+
+def find_met_in_model(model, mnx_id, compartment_id=None):
```
Midnighter: I think that letting this function raise a `KeyError` instead is more idiomatic.

scripts/generate_mnx_shortlists.py
```diff
@@ -0,0 +1,109 @@
...
+
+# Transpose the dataframe such that the index are now xref databases and the
+# column names are metanetx IDs.
+xref = xref.T
+
+# This is the current shortlist, expand this and rerun the script when adding
+# new code that requires the identification of key metabolites.
+# Included in the list below are energy metabolites, redox cofactors,
+# essential amino acids and some vitamins.
+shortlist = ['MNXM3', 'MNXM7', 'MNXM51', 'MNXM30', 'MNXM63', 'MNXM220',
```
Midnighter: I think that a better approach would be to store the identifiers that are currently in global variables (as done for energy couples and in other places) in YAML or CSV files. This script could then load those files and generate the shortlist from there. Seems more future proof to me.

scripts/generate_mnx_shortlists.py
```diff
@@ -0,0 +1,109 @@
...
+    xref : pandas.Dataframe
+        The metabolic model under investigation.
+
+    """
+    if ':' in xref:
+        return xref.split(':')[0]
+    elif 'MNX' in xref and ':' not in xref:
+        return 'mnx'
+
+
```
Midnighter: Would it make sense to limit the dataframe to identifiers from the shortlist before performing all these transformations?

ChristianLieven: It is a bit tricky because the initial dataframe isn't shaped in a way that allows that but it would make the execution of the script much faster.

I've made the appropriate changes to the script!

scripts/generate_mnx_shortlists.py
ChristianLieven: I'm not entire sure what you mean here and how to implement it for every single 'hardcoded' identifier. Since it is working quite well right now, I'd like to hold off on this somewhat larger change until we've published the tool. I do agree with you that handling these lists of hardcoded identifiers is important as soon as there are more than two people working on this, but I don't see this as a priority right now. I'd like to save it for the #GreatRefactoring2019 ;)

memote/support/consistency.py
ChristianLieven: Actually I noticed that the error check is no longer necessary as the boundaries are closed sensibly. I've made the appropriate change, and am now raising the error with your proposed method but I could not come up with a test case where this error would actually get raised and the optimisation would fail.

The amendments are in the next commit!

ChristianLieven:
@Midnighter I've responded to all you're comments and have pushed the corresponding corrections.

scripts/generate_mnx_shortlists.py
ChristianLieven: Opened this issue for us to keep track of it:
https://github.com/opencobra/memote/issues/324

PR #336, opened 2018-02-05
fix/sbo-import: replaced all instances of annotation.find_components_…

siddC:
…without_sbo_terms with sbo.find_components_without_sbo_terms prevent attribute error, which occured since the find_components_without_sbo_terms function was in memote/support/sbo.py instead of memote/support/annotation.py

* [ ] fix #x (issue number)
* [ ] tests added/passed

### DIFF
--- a/memote/suite/tests/test_sbo.py
+++ b/memote/suite/tests/test_sbo.py
@@ -37,7 +37,7 @@ def test_metabolite_sbo_presence(read_only_model):
     viewed here http://www.ebi.ac.uk/sbo/main/tree.
     """
     ann = test_metabolite_sbo_presence.annotation
-    ann["data"] = get_ids(annotation.find_components_without_sbo_terms(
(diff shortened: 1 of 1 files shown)


ChristianLieven:
Looking good Sidd,

I'm happy to merge once you've fixed flake8.

```memote/suite/tests/test_sbo.py:26:1: F401 'memote.support.annotation' imported but unused```

**Comment by Midnighter** [comment]
Thank you for the fixes @siddC. A note about git commit messages: We don't perfectly follow it but at least try to be in line with: https://seesparkbox.com/foundry/semantic_commit_messages

In short: Use a quick commit type identifier. Then phrase a short summary line (< 55 chars). After that write as much as you want. Everything should be in present tense and imperative voice, for example,

```
fix: correct the module import

Replace the import of module 'annotation' with 'sbo'.
```

PR #337, opened 2018-02-06
fix/sbo: prevent divide by zero error

siddC:
* [ ] fix #x (issue number)
* [ ] tests added/passed

Add try/except structure to `ann["metric"]` in `test_sink_specific_sbo_presence` to prevent ZeroDivisionError when no sink reactions are present. pytest skips this test instead.

### DIFF
--- a/memote/suite/tests/test_sbo.py
+++ b/memote/suite/tests/test_sbo.py
@@ -261,11 +261,11 @@ def test_sink_specific_sbo_presence(read_only_model):
     organism's compartments.
     """
     ann = test_sink_specific_sbo_presence.annotation
-    sink_react_list = helpers.find_sink_reactions(read_only_model)
+    sink_reactions = helpers.find_sink_reactions(read_only_model)
     ann["data"] = get_ids(sbo.check_component_for_specific_sbo_term(
-        sink_react_list, "SBO:0000632"))
+        sink_reactions, "SBO:0000632"))
     try:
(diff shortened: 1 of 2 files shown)


memote/suite/tests/test_sbo.py
```diff
@@ -261,8 +263,11 @@ def test_sink_specific_sbo_presence(read_only_model):
     ann = test_sink_specific_sbo_presence.annotation
     ann["data"] = get_ids(sbo.check_component_for_specific_sbo_term(
         helpers.find_sink_reactions(read_only_model), "SBO:0000632"))
-    ann["metric"] = len(ann["data"]) / len(
-        helpers.find_sink_reactions(read_only_model))
+    try:
+        ann["metric"] = len(ann["data"]) / len(
+            helpers.find_sink_reactions(read_only_model))
```
Midnighter: Since it's called multiple times (see above) can you please store the result of `find_sink_reactions` in a local variable.

memote/suite/tests/test_sbo.py
```diff
@@ -261,8 +263,11 @@ def test_sink_specific_sbo_presence(read_only_model):
...
+    except ZeroDivisionError:
+        pytest.skip()
```
Midnighter: What's missing here:
* Store a message `ann["message"] = "No sink reactions found.".
* Just in case set the metric to fully failed `ann["metric"] = 1.0.
* Set that message as the skip reason `pytest.skip(ann["message"])`.

PR #338, opened 2018-02-07
fix/basic: fix string-float type error in assert statement

siddC:
* [ ] fix #x (issue number)
* [ ] tests added/passed

Changed assert statement from comparing the length of `ann["metric"]` to just comparing ann["metric"] itself, preventing a type error (as floats don't have a length)


Midnighter:
Thank you for all the fixes!

PR #350, opened 2018-02-14
Matrix tests

### LINKED ISSUE #233: test if (or how much) the stoichiometric matrix is ill-conditioned
As pointed out by Steffen Klamt in the google document:

> "I find always useful is the absolute maximum and the absolute (non-zero) minimum of coefficients in the stoichiometric matrix since they tell us something about matrix conditioning and may point to potential numerical issues."

I think this is a very relevant point. I have also often encountered problems with numerical issues that were associated with ill-conditioned matrices (most times associated with the biomass coefficients). 

The condition number can be calculated in different ways:

https://en.wikipedia.org/wiki/Condition_number

For this particular purpose, I think the absolute ratio between the largest and smallest coefficients should be sufficient. 

For instance, for iJO1366 the result is 2.7e7. From my (limited) experience, something above 1e9 should raise some flags...

siddC:
* [ ] fix #233 
* [ ] tests added/passed

### Summary
These commits add 4 functions to `matrix.py` and which are tested in `test_matrix.py` by 4 other functions. These functions test for:
* number of independent conservation relations/conservation pools
* number of steady-state flux vectors
* matrix rank
* degrees of freedom

### Discussion

The test for `number of independent conservation relations/conservation pools` comes from listing the dimension of the left nullspace, it does not actually list the conserved moieties themselves. As Niko and James have stated, the left nullspace of large S matrices are not easily interpretable (iJO1366 was tested in this case). For now, this issue of listing the actual conservation pools has been shelved, and may be something we look at in the future.

Degrees of freedom (DOF) in this case calculates the DOF of the S-Matrix, i.e. the number of "free variables". This ends up being equivalent to the dimension of the right nullspace, and is calculated as such. From the comments Steffen Klamt posted, it seemed like this is the statistic he was referring to, but if this is not the case, then we will need to correct it in the future.

Besides these two issues, everything else should work as intended. I could not think of any more statistics to add for the S-Matrix, but if anyone has any more ideas, then let me know and I will try to add them in a later commit.


memote/support/matrix.py
```diff
@@ -0,0 +1,93 @@
...
+    )
+    abs_matrix, _, _ = abs(s_matrix)
+
+    absolute_max_coef = abs_matrix.max()
+    absolute_non_zero_min_coef = np.min(abs_matrix[np.nonzero(abs_matrix)])
+
+    return (absolute_max_coef, absolute_non_zero_min_coef)
+
+
+def number_independent_conservation_relations(model):
```
ChristianLieven: Given by the name I'd expect this function to return an integer of the number of independent conservation relations and not the left nullspace matrix itself.

It generally makes more sense to put as much of the logic into these helper functions and only pass the final result to the tests. This is because we currently don't unit test the tests themselves.

Also see my comment above in the corresponding test (`test_number_independent_conservation_relations`).

ChristianLieven: The thing is that currently `data` can only be displayed on the reports if it is one of the following (quoted from `annotate` docstring in `utils.py`):

 - 'number' : 'data' is a single number which can be an integer or
          float and should be represented as such.
 - 'count' : 'data' is a list, set or tuple. Choosing 'count' will
          display the length of that list e.g. number of metabolites without formula.
 - 'percent' : Instead of 'data' the content of 'metric' ought to be
          displayed e.g. percentage of metabolites without charge.
          'metric' is expected to be a floating point number.
- 'raw' : 'data' is ought to be displayed "as is" without formatting.
          This option is appropriate for single strings or a boolean output.

memote/support/matrix.py
```diff
@@ -0,0 +1,93 @@
...
+    """Return the amount of conserved metabolic pools."""
+    s_matrix, _, _ = con_helpers.stoichiometry_matrix(
+        model.metabolites, model.reactions
+    )
+    ln_matrix = con_helpers.nullspace_basis(s_matrix.T)
+    return ln_matrix
+
+
+def number_steady_state_flux_solutions(model):
```
ChristianLieven: Same as comment above. I'd expect an integer not a matrix as the return value.

memote/suite/tests/test_matrix.py
```diff
@@ -0,0 +1,124 @@
...
+
+    This test will return the number of conservation relations, i.e.
+    conservation pools through the left null space of the S-Matrix.
+
+    This test is not scored, as the dimension of the left null space
+    depends on the S-Matrix constructed, which is system-specific.
+    """
+    ann = test_number_independent_conservation_relations.annotation
+    ann["data"] = matrix.number_independent_conservation_relations(model)
+    ann["metric"] = ann["data"].shape[1]
```
ChristianLieven: I think this operation should probably take part in the helper function itself:
```ann["data"].shape[1]```

In other words I think a test named `number_independent_conservation_relations` should probably return an integer and not a matrix. If we need the matrix somewhere else we should probably split up the helper functions into one that generates the left nullspace matrix and one that extracts the corresponding number of relations from it?

Moreover, `ann["metric"]` should only be assigned if there is a ratio/ percentage worth showing or the test itself is scored. In this case, I'd leave it unassigned. Take a look at one of my comments below where I quote the docstring from the `annotate` function in `utils.py`. Perhaps that explains a bit better what `ann["data"]` can and cannot be.

siddC:
Closing this PR and sending a new one to fix travis ci and bad diff issues.

PR #351, opened 2018-02-15
fix: functions return objects not str ids

siddC:
* [ ] fix #x (issue number)
* [ ] tests added/passed

### Summary
* `find_metabolites_produced_with_closed_bounds`
* `find_metabolites_consumed_with_closed_bounds`

These two functions were returning a list of metabolite ids, not metabolite objects. To keep memote functions consistent, these functions now return a list of metabolite objects, not string ids, as previously discussed with Moritz and Christian. An entry to `History.rst` will be added after #350 is pulled so as to not create unnecessary merge conflicts within `History.rst`


ChristianLieven:
Don't worry about the merge conflicts in History.rst that's an easy fix for me. However, take a look at why Travis might be failing!

PR #354, opened 2018-02-15
Matrix tests

### LINKED ISSUE #233: test if (or how much) the stoichiometric matrix is ill-conditioned

siddC:
* [x] fix #233 
* [x] tests added/passed

### Summary

These changes should address the concerns in issue #233 except for adding in/testing for a list of parallel routes. Additionally, the number of independent conserved metabolite relations, i.e. conservation pools, is added but the actual pools themselves are not shown as the left-nullspace of large network reconstructions is extremely difficult to interpret.

### DIFF
--- a/memote/support/matrix.py
+++ b/memote/support/matrix.py
@@ -74,15 +74,15 @@ def degrees_of_freedom(model):
     ------
     This specifically refers to the dimensionality of the right nullspace
     of the S matrix, as dim(Null(S)) corresponds directly to the number of
-    free variables in the system [1]. The forumla used calculates this using 
+    free variables in the system [1]. The forumla used calculates this using
     the rank-nullity theorem [2].
 
     References:
     -----------
-    .. [1] Fukuda, K. & Terlaky, T. Criss-cross methods: A fresh view on 
+    .. [1] Fukuda, K. & Terlaky, T. Criss-cross methods: A fresh view on
            pivot algorithms. Mathematical Programming 79, 369-395 (1997).
 
-    .. [2] Alama, J. The Rank+Nullity Theorem. Formalized Mathematics 15, 
+    .. [2] Alama, J. The Rank+Nullity Theorem. Formalized Mathematics 15,
            (2007).
 
     """
--- a/memote/support/matrix.py
+++ b/memote/support/matrix.py
@@ -80,10 +80,10 @@ def degrees_of_freedom(model):
     References:
     -----------
     .. [1] Fukuda, K. & Terlaky, T. Criss-cross methods: A fresh view on 
-       pivot algorithms. Mathematical Programming 79, 369-395 (1997).
+           pivot algorithms. Mathematical Programming 79, 369-395 (1997).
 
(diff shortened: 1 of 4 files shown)


memote/support/matrix.py
```diff
@@ -0,0 +1,93 @@
...
+LOGGER = logging.getLogger(__name__)
+
+
+def absolute_extreme_coefficient_ratio(model):
+    """Return the absolute max and absolute non-zero min coefficients."""
+    # S-Matrix with absolute values:
+    s_matrix, _, _ = con_helpers.stoichiometry_matrix(
+        model.metabolites, model.reactions
+    )
+    abs_matrix, _, _ = abs(s_matrix)
```
Midnighter: `s_matrix` is a 2D `ndarray` so I don't understand why its absolute matrix should return a triple.

memote/support/matrix.py
```diff
@@ -0,0 +1,93 @@
...
+
+    absolute_max_coef = abs_matrix.max()
+    absolute_non_zero_min_coef = np.min(abs_matrix[np.nonzero(abs_matrix)])
```
Midnighter: Since you're dealing with the absolute values `abs_matrix[abs_matrix > 0].min()` should be fine.

memote/support/matrix.py
```diff
@@ -0,0 +1,93 @@
...
+def degrees_of_freedom(model):
+    """
+    Return the degrees of freedom, i.e. number of "free variables".
+
+    This specifically refers to the dimensionality of the right nullspace
+    of the S matrix, as dim(Null(S)) corresponds directly to the number of
+    free variables in the system. The forumla used calculates this using the
+    rank-nullity theorem. For more information, see the links below.
+
+    See Also:
```
Midnighter: `See Also` is reserved for pointers to other related functions in the package. Please use a `References` section here (cf. https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#sections point 11).

siddC:
I've changed the logic in `absolute_extreme_coefficient_ratio` with regards to min, max, and absolute value. I've also fixed the docstring for `degrees_of_freedom` and added in the test changes into `test_config.yml` and in `HISTORY.rst`. Let me know if anymore changes need to be made to this branch.

Midnighter:
Nice work @siddC! I like your commit structuring and their messages.

PR #365, opened 2018-03-06
Fix/test description

### LINKED ISSUE #331: Possibly remove the first line of the docstring from the test 'description' in the report
> I think the problem is the wording of all the description. They should be description of the test, not expectation of the result. If we are talking about the result, then clearer/stronger words should be used. MUST/SHOULD, REQUIRED etc.
- Nicolas Le Novere

> I understand where you're coming from Nicolas. The reason for the rather specific way that these descriptions here are written (imperative, short and terse) is that they are in-fact the first lines the our docstrings of the corresponding test functions in our code. As such they are subject to several good-practises on the implementation side (numpy/ scipy docstring definition, pep8, pep257 http://blog.dolphm.com/pep257-good-python-docstrings-by-example/). 
For all of our test cases the descriptions are hardcoded within the function docstrings as this allows us to automatically propagate this into the report and the official memote documentation. 
We have taken to including the summary lines in the updated report that we are currently working on, but given your valid criticism we will reconsider including the very first line (the one required to be in imperative mood), and take to rewording the summaries to be more descriptive.
Thank you!
- My response

siddC:
* [x] fix #331 
* [x] tests added/passed

While the first line in a function docstring is important for describing expected function behavior and output, it is not useful information for an end user, who would rather see a summary of the test being run on the model and its significance. Therefore, the first line of a docstring is now trimmed from the function summary that is shown to end users. This involved generating a small trimming function in `memote/utils.py` and replacing the summary from `func.__doc__` to `trim_description(func)`.


Midnighter:
Hey, @siddC, I think it'd be more flexible if you use functionality from `numpydoc` directly. They have an API to easily extract different sections of the docstrings.

siddC:
Running `tox -e py36` I found that the build is failing because all the `test_for_` files have the following error:

```E   ModuleNotFoundError: No module named 'importlib_resources'```

I haven't touched these files at all so I'm not sure why this is the case

Midnighter:
Can you run `tox -e py36 --recreate`? Or maybe `importlib-resources` is missing from the `requirements.txt`?

siddC:
`importlib-resources` is in the `requirements.txt` file. Running `tox -e py36 --recreate` outputted:

```4 failed, 278 passed, 1 skipped, 2 xfailed, 2 warnings in 22.23 seconds```

All 4 errors occur in `test_for_utils.py`:

```tests/test_for_utils.py:99: AssertionError```

siddC:
Figured out the problem. Line 99 in `test_for_utils.py` is: 
```assert res.annotation["summary"] == summary```

But since we don't want that anymore, we'll need to change this as well. I'll take a look into `numpydoc` and see if I can use it instead.

siddC:
Hey @Midnighter, looking at the [this](https://pypi.python.org/pypi/numpydoc?) and [this](https://numpydoc.readthedocs.io/en/latest/), I'm not sure what you mean by the API. While there are some functions that can produce plot directives in the Example sections of functions that use matplotlib and such, it doesn't seem like there is anything that can simply trim off the first line in a docstring. I might be missing something, but I don't think I am and it seems simpler to just use a 3 line `utils.py` function for ouputting the summary to end users than using `numpydoc`. All that is needed right now is to change the test in `test_for_utils.py` and this issue should be resolved.

What are your thoughts on this?

Midnighter:
@ChristianLieven you used functions from `numpydoc` before. Do you remember how?

Midnighter:
Actually, take a look at [SO](https://stackoverflow.com/questions/37929851/how-can-i-parse-a-numpydoc-docstring-and-access-components#41083968) @siddC.

Midnighter:
This would allow use to add other sections to the docstrings (if we ever want to do that) and only ever extract a user friendly "Extended Summary".

siddC:
Hmm, it wouldn't change the existing code too much and I think I can implement it quickly. I'll test it out and see how it goes. However, would this add `numpydoc` as another dependency to memote?

Midnighter:
Yes, indeed. It needs to be added to `./requirements.txt` and to the list `PKG_INFO` in `memote/version_info.py`.

siddC:
Using numpydoc, I was able to change the summary so it only outputted the Extended Summary and not the first line. However, even after adding `numpydoc` to `./requirements.txt` and in `PKG_ORDER` in `memote/version_info.py`, `tox -e py36` still failed with the error:

```E   ModuleNotFoundError: No module named 'numpydoc'```

Midnighter:
Did you try to `--recreate` again?

siddC:
Its currently running. Waiting for it to finish

siddC:
It just finished and passed.

ChristianLieven:
Looks good to me! Sorry for not reacting earlier, I was in meetings most of today.

PR #368, opened 2018-03-08
Feat/better error handling

ChristianLieven:

#### Description:
Tests that are skipped will always have an empty (`none`) data attribute. For tests which pass or fail this may be the case:
- when the user writing the tests does not to set the data attribute,
- or the helper function providing the data throws an exception.

When the display type `count` is chosen, the report attempts to determine the length of whatever is stored as data. When data is `null` (Angular equivalent of the python `none`) this results in an error which breaks the interface.

Since either of those cases above can be seen as errors (by a user or by one of the helper functions), the fix implemented in this PR consists of setting the `result` attribute of each test from passed or failed to error.

### DIFF
--- a/memote-report-app/src/app/report-data.service.ts
+++ b/memote-report-app/src/app/report-data.service.ts
@@ -17,8 +17,8 @@ export class ReportDataService {
   public loadResults(): void {
     // TODO: Might want to parse and decompress a string in future.
     // const data = JSON.parse((<any>window).data);
-    this.convertResults((<any>window).data);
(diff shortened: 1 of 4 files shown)


memote-report-app/src/app/report-data.service.ts
```diff
@@ -39,7 +39,7 @@ export class ReportDataService {
   public isScored(string) {
     return this.scoredTests.includes(string);
   }
-
+
```
kvikshaug: No need to add whitespace here, you might want to configure your editor/IDE to automatically trim trailing whitespace

PR #369, opened 2018-03-08
Fix/optimize sbc

### LINKED ISSUE #104: Check for thermodynamically infeasible loops
### Description

Memote currently lacks tests to identify reactions that are involved in Type III loops i.e. reactions that carry flux even though all boundary reactions are closed.

### Implementation

FVA with blocked exchanges if the min/max fluxes == lb/ub => rxn is involved in a loop

siddC:
* [x] fix #104 
* [x] tests added/passed

Performing `test_find_stoichiometrically_balanced_cycles` currently is intractable for large models. These commits change the `find_stoichiometrically_balanced_cycles` function to work faster, by replacing 2 FVA calculations with only 1.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,7 @@ History
 
 Next Release
 ------------
+
 * Test summary only displays extended narrative summary describing test,
   and not one-line summary describing expected function behavior/output
 * Fix the following bugs:
@@ -17,12 +18,12 @@ Next Release
 * Add filter in ``report_data_service`` that changed the test result status to
   "error" when the data attribute is ``null``, thus avoiding that the report
   interface breaks when trying to access data.
+* Add test for identifying stoichiometrically balanced cycles in models
 
 0.6.1 (2018-03-01)
 ------------------
+
 * Emergency fix for distributing required JSON file.
-
-
 
 0.6.0 (2018-02-27)
 ------------------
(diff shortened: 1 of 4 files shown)


ChristianLieven:
>This is a nice attempt at speeding up the test, however, it would not identify irreversible loops since you're only reporting the union of reactions with flux in {-1, 1}.

What do you mean by irreversible loop? If it is something like A->B and B->A then the way Sidd is doing it currently should catch that or not?

> Unless there's another way to test for this, I think the way to go forward would simply be to set a timeout on the solver.

I think the current approach is a massive improvement over the previous one and very much in line with what has been done previously in the COBRA Toolbox for instance:
https://github.com/opencobra/cobratoolbox/blob/b80de23adff2b01e4fb468c53cc8ceef2a8203dc/src/analysis/topology/testForTypeIIIPathways.m

Midnighter:
I mean a single irreversible loop like A -> B -> C. In an FVA it would show up as max: 1, 1, 1 and min: 0, 0, 0. So not caught by the test.

ChristianLieven:
But how is that a loop?

Midnighter:
Sorry @siddC, I had a logic error. Thinking about intersection when you clearly used the union :grimacing: My bad. The last point still stands, though, exception catching is not necessary.

siddC:
So in summary, the only change that needs to be made is to remove the exception handling, correct? =

Midnighter:
I know you didn't introduce it but yes, please remove the exception handling. And in my opinion since we are converting reaction objects back to IDs anyhow, you can just shorten the last few lines to:

```python
    return ((fva_result["maximum"] == 1) | (fva_result["minimum"] == -1)).index
```

It might have to be turned into a native Python `list`.

siddC:
Okay, I've removed the exception handling and simplified the return statement

siddC:
I've reverted back to the original return statement (without the exception handling) and the `tox -e py36` has passed so I'll just leave it as is

PR #371, opened 2018-03-12
Feat/report fixed pure metabolic reactions

### LINKED ISSUE #333: Identify and report any purely metabolic reactions which have fixed constraints
Possibly look for internal (non-boundary) reactions that are constrained and report those!

This would just help to get an idea what state the model is in/ whether the model builders have made certain assumptions.

siddC:
* [x] fix #333 
* [x] tests added/passed

Add a new function `find_constrained_pure_metabolic_reactions` to address #333 inside `basic.py`. Added `test_find_constrained_pure_metabolic_reactions` to `test_basic.py` and registered it in `memote/suite/templates/test_config.yml`. Also added unit test for the base function (also called `test_find_constrained_pure_metabolic_reactions`) in `test_for_basic.py`.

`find_constrained_pure_metabolic_reactions` simply looks through the set of all purely metabolic reactions and returns those which have a lower bound > 0. The only reaction found when testing on iJO1366 was on ATPM. However, this test is very much system-specific, so it has no mandatory 'pass' criteria.

### DIFF
--- a/memote/support/basic.py
+++ b/memote/support/basic.py
@@ -221,47 +221,14 @@ def find_pure_metabolic_reactions(model):
     return set(model.reactions) - (exchanges | transporters | biomass)
 
 
-def find_constrained_pure_metabolic_reactions(model_or_rxns):
-    """
-    Return purely metabolic reactions with fixed constraints.
-
-    Purely metabolic reactions in this case are defined as reactions that are
-    neither transporters, exchanges, nor pseudo.
-
-    Parameters
-    ----------
-    model_or_rxns : cobra.Model or set of cobra.Reaction
-        The metabolic model under investigation or a set of reactions from
-        the metabolic model under investigation.
-
-    """
-    pmr = None
-    if isinstance(model_or_rxns, cobra.core.Model):
-        pmr = find_pure_metabolic_reactions(model_or_rxns)
-    elif isinstance(model_or_rxns, set) and isinstance(list(model_or_rxns)[0],
-                                                       cobra.core.Reaction):
-        pmr = model_or_rxns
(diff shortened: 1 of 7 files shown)


memote/support/basic.py
```diff
@@ -220,6 +220,16 @@ def find_pure_metabolic_reactions(model):
     return set(model.reactions) - (exchanges | transporters | biomass)
 
 
+def is_constrained_reaction(rxn):
+    """Return whether a reaction has fixed constraints."""
+    if rxn.reversibility:
+        return True if (
```
Midnighter: You can simply return the evaluated expression ;)
```python
return (rxn.lower_bound > -1000 or rxn.upper_bound < 1000)
```

memote/support/basic.py
```diff
@@ -220,6 +220,16 @@ def find_pure_metabolic_reactions(model):
...
+            rxn.lower_bound != -1000 or rxn.upper_bound != 1000) else False
+    else:
+        return True if (
+            rxn.lower_bound != 0 or rxn.upper_bound != 1000) else False
```
Midnighter: Same as above
```python
return (rxn.lower_bound > 0 or rxn.upper_bound < 1000)
```

memote/support/basic.py
siddC: Ah you're right! I'll have it fixed first thing tomorrow morning!

Midnighter:
I just realized that for these two test cases it'd actually be nice to not only store the reaction ID in `ann["data"]` but a triple: `(rxn.id, rxn.lower_bound, rxn.upper_bound)`. Quick fix to do with those changes above. And then I guess it's really, really done:tm:  :stuck_out_tongue_winking_eye:

PR #396, opened 2018-03-19
Fix/find direct metabolites

### LINKED ISSUE #257: Test for a low ratio of direct metabolites in the biomass reaction
#### Problem description

> Some biomass precursors are taken from the media and directly consumed by the biomass reaction. It might not be a problem for some ions or metabolites with known auxotrophy, but when there are too many of these metabolites, it points to artefacts of gap-filling. This mostly happens in case of automatically-generated models. Starting with a complicated automatically generated biomass formulation followed by gap-filling (which basically tries to minimise the added reactions), would lead to many biomass metabolites being "direct mets". It might be useful to report the ratio of biomass metabolites that are direct mets.

- Parizad Babayi, comment on the manuscript.

siddC:
* [x] fix #257 
* [x] tests added/passed

This fix takes into account the flux of `tra_bou_bio` reactions and uses them to detect and remove false positives from the original function. Also changes the unit tests to account for this.


memote/support/biomass.py
```diff
@@ -176,14 +180,56 @@ def find_direct_metabolites(model, reaction):
...
+    main_comp = helpers.find_compartment_id_in_model(model, 'c')
+    exc_space = helpers.find_compartment_id_in_model(model, 'e')
+    for met in tra_bou_bio_mets:
+        for rxn in met.reactions:
+            not_biomass_rxn = rxn not in biomass_rxns
+            extracellular = met.compartment == exc_space
+            # if reactant metabolite in "e"
+            if met in rxn.reactants and not_biomass_rxn and extracellular:
+                product_comps = set([p.compartment for p in rxn.products])
+                # e reactant, no product (outward flux)
```
ChristianLieven: Just fyi, and I'm certainly also guilty of this, but great comments go a long way in making your code accessible to others and your future self. These are somewhat cryptic and perhaps only clear to you once you've fully understood the code, at which point they become redundant :D

memote/support/biomass.py
```diff
@@ -176,14 +180,56 @@ def find_direct_metabolites(model, reaction):
...
+                if len(reactant_comps) == 0:
+                    flux_sum[met] += float(
+                        tra_bou_bio_fluxes[get_ids([rxn])])
+                # e product, c reactant (outward flux)
+                elif main_comp in reactant_comps:
+                    flux_sum[met] += -float(
+                        tra_bou_bio_fluxes[get_ids([rxn])])
+            # for all non-e metabolites
+            elif not_biomass_rxn and not extracellular:
+                flux_sum[met] += np.abs(
```
ChristianLieven: According to CodeCov: https://codecov.io/gh/opencobra/memote/compare/f7dbca89ddb5406b2ba4566a88e341c16ad5c051...ab15e59a0311be476099ce7181ff35bfc413cd64/diff

The control flow statements from lines 206 to 225 are not covered in any tests (which is also why the build fails)

memote/support/biomass.py
```diff
@@ -176,14 +180,56 @@ def find_direct_metabolites(model, reaction):
...
+                            for rxn in met.reactions
+                            if rxn not in biomass_rxns])
+
+    solution = model.optimize()
+    if solution.objective_value == 0:
+        raise ValueError("objective value is zero. Model will not grow.")
 
-    return [met for met in precursors if met.reactions.issubset(combined_set)]
+    tra_bou_bio_fluxes = solution.fluxes[get_ids(rxns_of_interest)]
+    flux_sum = pd.DataFrame(index=tra_bou_bio_mets, columns=["sum"], data=0).T
```
ChristianLieven: I have a feeling that using a pandas data frame here is a bit of an overkill and a simple dictionary could be just as effective since you're essentially only needing the `tra_bou_bio_mets` as keys and the corresponding flux sums as values.

memote/support/helpers.py
```diff
@@ -367,6 +367,31 @@ def find_exchange_rxns(model):
...
+        A cobrapy metabolic model
+
+    biomass : list or cobra.Reaction, optional
+        A list of cobrapy biomass reactions
+
+    """
+    # exchanges in this case also refer to sink and demand reactions
+    exchanges = set(model.exchanges)
+    transporters = set(find_transport_reactions(model))
+    if biomass is None:
```
ChristianLieven: Please see my comment above, I don't fully understand why this is necessary.

tests/test_for_support/test_for_biomass.py
```diff
@@ -237,6 +237,58 @@ def direct_met_single_compartment(base):
...
+    rxn1 = cobra.Reaction("MET_Atec", lower_bound=-1000, upper_bound=1000)
+    rxn1.add_metabolites({met_a: 1, met_a1: -1})
+    rxn2 = cobra.Reaction("MET_Btec", lower_bound=-1000, upper_bound=1000)
+    rxn2.add_metabolites({met_b: 1, met_b1: -1})
+    rxn3 = cobra.Reaction("MET_Ctec", lower_bound=-1000, upper_bound=1000)
+    rxn3.add_metabolites({met_c: 1, met_c1: -1})
+    a2b = cobra.Reaction("A2B", lower_bound=-1000, upper_bound=1000)
+    a2b.add_metabolites({met_a: 1, met_b: -1})
+    base.add_reactions([rxn, rxn1, rxn2, rxn3, a2b])
+    base.add_boundary(met_a1, ub=5)
```
ChristianLieven: By adding this exchange reaction you've turned the otherwise correct test case for a false positive into just a regular test again. Remove this boundary reaction to cover the false positive edge case.

I'm a bit confused, that you're asserting the test with this to come up with zero direct metabolites. The way I see it all of these are direct metabolites as they all appear only in transport, boundary and the biomass reaction.

I have a feeling that adding formulae to the metabolites is necessary for this test to work.

memote/support/biomass.py
```diff
@@ -176,14 +180,56 @@ def find_direct_metabolites(model, reaction):
         that are taken up to be consumed by the biomass reaction only.
 
     """
-    transport_reactions = set(helpers.find_transport_reactions(model))
-    exchange_reactions = set(model.exchanges)
-    biomass_reactions = set(helpers.find_biomass_reaction(model))
-
-    combined_set = transport_reactions | exchange_reactions | biomass_reactions
+    biomass_rxns = set(helpers.find_biomass_reaction(model))
+    tra_bou_bio_rxns = helpers.find_tra_bou_bio_reactions(model, biomass_rxns)
```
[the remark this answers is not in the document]
Midnighter: That's a terrible function name btw :stuck_out_tongue:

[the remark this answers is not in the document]
ChristianLieven: Transport/Boundary/Biomass reactions :D I thought it was brilliant!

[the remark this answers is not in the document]
siddC: The reason I passed in the argument was to remove unnecessary/redundant calculations. Both this function and the function it called (`find_tra_bou_bio_reactions`) call the same function (`find_biomass_reaction`). So in order to prevent the function being called twice, I added in an optional, second parameter so the biomass reaction would only be calculated once.

memote/suite/tests/test_biomass.py
```diff
@@ -278,19 +278,23 @@ def test_direct_metabolites_in_biomass(model, reaction_id):
...
+        ann["metric"][reaction_id] = 1.0
+        ann["message"][reaction_id] = "This model does not grow."
+        pytest.skip(ann["message"])
+    else:
+        ann["metric"][reaction_id] = len(ann["data"][reaction_id]) / \
+            len(biomass.find_biomass_precursors(model, reaction))
+        ann["message"][reaction_id] = wrapper.fill(
+            """{} contains a total of {} direct metabolites ({:.2%}).
+            Specifically these are: {}.""".format(
+                reaction_id, len(ann["data"][reaction_id]),
```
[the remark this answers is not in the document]
siddC: The reason I had it like this was to ensure nothing went over 79 characters. I can try and move stuff around and see if I make it more readable.

memote/support/biomass.py
```diff
@@ -176,14 +180,56 @@ def find_direct_metabolites(model, reaction):
...
     precursors = find_biomass_precursors(model, reaction)
+    tra_bou_bio_mets = [met for met in precursors if
+                        met.reactions.issubset(tra_bou_bio_rxns)]
+    rxns_of_interest = set([rxn for met in tra_bou_bio_mets
+                            for rxn in met.reactions
+                            if rxn not in biomass_rxns])
+
+    solution = model.optimize()
+    if solution.objective_value == 0:
+        raise ValueError("objective value is zero. Model will not grow.")
```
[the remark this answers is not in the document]
siddC: I was going to ask about the logging thing. But yes, I can add this in and fix it up using `consistency.py` as a template

tests/test_for_support/test_for_biomass.py
siddC: Fixed the test case, but now it's pretty much identical to the `only_direct_mets` test model. Should I delete one of them?

memote/suite/tests/test_biomass.py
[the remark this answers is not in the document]
siddC: Fixed!

HISTORY.rst
```diff
@@ -14,6 +14,7 @@ Next Release
 * Allow tests and test module to be skipped or run exclusively when creating
   a snapshot report.
 * Fix some typos
+* find_direct_metabolites now detects and removes false_positives
```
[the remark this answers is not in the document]
siddC: Fixed!

tests/test_for_support/test_for_biomass.py
siddC: Ah I see the issue now. The function is working as intended as the flux between `lipid_c` and `lipid_e` is zero, which is why it's considered a direct metabolite. I will fix the model so it does detect as a false positive and is not counted as a direct metabolite

memote/support/biomass.py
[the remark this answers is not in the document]
siddC: Added!

memote/support/biomass.py
siddC: Okay, I've tried clarifying the code!

memote/support/biomass.py
```diff
@@ -176,14 +192,83 @@ def find_direct_metabolites(model, reaction):
...
-    precursors = find_biomass_precursors(model, reaction)
-
-    return [met for met in precursors if met.reactions.issubset(combined_set)]
+    biomass_rxns = set(helpers.find_biomass_reaction(model))
+    tra_bou_bio_rxns = helpers.find_tra_bou_bio_reactions(model, biomass_rxns)
+    try:
+        precursors = find_biomass_precursors(model, reaction)
+        main_comp = helpers.find_compartment_id_in_model(model, 'c')
+        ext_space = helpers.find_compartment_id_in_model(model, 'e')
+    except KeyError as k:
```
Midnighter: Other than in list comprehensions single letter variables are bad style. It'd be better to use `as err` or even `as error` here.

memote/support/biomass.py
```diff
@@ -176,14 +192,83 @@ def find_direct_metabolites(model, reaction):
...
+        LOGGER.error("Failed to properly identify cytosolic and extracellular "
+                     "compartments.")
+        raise KeyError("The cytosolic and/or extracellular compartments could "
+                       "not be identified. "
+                       "find_compartment_id_in_model: " + str(k))
```
Midnighter: If you are raising another error here I don't fully understand why you are catching the `KeyError` in the first place. Or maybe you mean to raise this error with the traceback of the previous error? See `raise_with_traceback` in http://python-future.org/compatible_idioms.html#raising-exceptions

memote/support/biomass.py
```diff
@@ -176,14 +192,83 @@ def find_direct_metabolites(model, reaction):
...
+                elif main_comp in reactant_comps:
+                    flux_sum[met] += -float(
+                        tra_bou_bio_fluxes[get_ids([rxn])])
+            # for all non-e metabolites (cannot be false positives, flux is
+            # automatically non-negative)
+            elif not_biomass_rxn and not extracellular:
+                flux_sum[met] += np.abs(
+                    float(tra_bou_bio_fluxes[get_ids([rxn])]))
+
+    return [met for met in flux_sum.T.loc[flux_sum.T['sum'] > 0].index]
```
Midnighter: Albeit nicely commented, this function is really quite long. Let's take a look tomorrow whether we can break it down into smaller logical units.

memote/support/helpers.py
```diff
@@ -367,6 +367,31 @@ def find_exchange_rxns(model):
             if extracellular in rxn.get_compartments()]
 
 
+def find_tra_bou_bio_reactions(model, biomass=None):
```
Midnighter: As I mentioned I'm really not a fan of abbreviated function and argument names. Internally, that's totally fine with me but externally not so much.

tests/test_for_support/test_for_biomass.py
```diff
@@ -370,8 +566,13 @@ def test_fast_growth_default(model, boolean):
...
 @pytest.mark.parametrize("model, number", [
-    ("direct_met_single_compartment", 1),
-    ("precursors_producing", 0)
+    ("only_direct_mets", 3),
+    ("only_direct_mets_false_positive_candidate_EX_reactant_T_product", 3),
+    ("only_direct_mets_false_positive_candidate_EX_reactant_T_reactant", 3),
+    ("only_direct_mets_false_positive_candidate_EX_product_T_reactant", 3),
+    ("only_direct_mets_false_positive_candidate_EX_product_T_product", 3),
+    ("direct_mets_with_false_positive", 2),
+    ("precursors_producing", 0),
```
Midnighter: Very extensive tests, I like it :+1:

tests/test_for_support/test_for_biomass.py
```diff
@@ -380,6 +581,21 @@ def test_find_direct_metabolites(model, number):
         assert len(biomass.find_direct_metabolites(model, rxn)) is number
 
 
+@pytest.mark.parametrize("model", [
+    pytest.param("direct_met_no_growth",
```
Midnighter: You seem to be mixing different things here. Since we use `pytest-raises`, you can simply say:
```python
[
pytest.mark.raises("direct_met_no_growth", exception=ValueError),
...
]
```

memote/support/biomass.py
siddC: Got it! I'll fix this in the next commit!

memote/support/biomass.py
siddC: Yeah, exactly. I wanted to catch the exception, since finding the proper `c` and `e` compartments was necessary for detecting false positives and realized that models without properly labeled compartments would raise these exceptions, so I wanted to catch them and show the full traceback. I'll change this to `raise_with_traceback`

memote/support/helpers.py
siddC: Yeah, that was the name Christian suggested, so I just stuck with it. I can change it to something else

tests/test_for_support/test_for_biomass.py
siddC: For this, I was simply following the template in `test_for_consistency.py`, which uses the same basic format to catch errors (as seen in line 740):

~~~
@pytest.mark.parametrize("model", [
    pytest.param("missing_energy_partner",
                 marks=pytest.mark.raises(exception=ZeroDivisionError)),
    pytest.param("infeasible",
                 marks=pytest.mark.raises(exception=Infeasible))
], indirect=["model"])
def test_find_reactions_with_unbounded_flux_default_condition_errors(model):
    """Expect the number of unbounded and blocked metabolites to be correct."""
    consistency.find_reactions_with_unbounded_flux_default_condition(model)
~~~

tests/test_for_support/test_for_biomass.py
siddC: I had to put many of these in to pass the Codecov tests

memote/support/helpers.py
siddC: Copying here to explain why I did this:

>The reason I passed in the argument was to remove unnecessary/redundant calculations. Both this function and the function it called (find_tra_bou_bio_reactions) call the same function (find_biomass_reaction). So in order to prevent the function being called twice, I added in an optional, second parameter so the biomass reaction would only be calculated once.

P.S. `find_tra_bou_bio_reactions` has been renamed to `find_interchange_biomass_reactions`

PR #399, opened 2018-03-23
fix: get_ids_and_bounds now returns list of str

siddC:
* [ ] tests added/passed

`get_ids_and_bounds` used to return a list of tuples, which caused errors with the `truncate` function in `util.py`. It has been fixed to return a list of str, removing the `TypeError` exceptions being raised when running `memote run`

### DIFF
--- a/memote/utils.py
+++ b/memote/utils.py
@@ -127,7 +127,8 @@ def get_ids(iterable):
 
 def get_ids_and_bounds(iterable):
     """Retrieve the identifier and bounds of a  number of objects."""
-    return [(elem.id, elem.lower_bound, elem.upper_bound) for elem in iterable]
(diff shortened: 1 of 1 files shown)


memote/utils.py
```diff
@@ -127,7 +127,8 @@ def get_ids(iterable):
 
 def get_ids_and_bounds(iterable):
     """Retrieve the identifier and bounds of a  number of objects."""
-    return [(elem.id, elem.lower_bound, elem.upper_bound) for elem in iterable]
+    return [elem.lower_bound + "< " + elem.id + "< " + elem.upper_bound for
```
Midnighter: Sorry, since it should be `<=` anyhow, please use the format function, i.e.,
```python
return ["{0.lower_bound} <= {0.id} <= {0.upper_bound}".format(elem) for elem in iterable
```

Issue #384, opened 2018-03-14
Transport reactions not detected


lenov:
Memote tells me there are no transport reactions in my model. According to memote's documentation those reactions are detected based on the formula of the compounds being the same on both sides of the reaction, but in different compartments. Here is a bit of my SBML file describing the transport of oxygen between mitochondrion and cytosol. Should it not be detected?

      <species boundaryCondition="false" compartment="c" constant="false" fbc:charge="0" fbc:chemicalFormula="O2" hasOnlySubstanceUnits="false" id="M_o2_c" initialConcentration="0" metaid="_8da2f742-2592-4760-987a-e76067d9c0cf" name="dioxygen">
        <notes>
          <body xmlns="http://www.w3.org/1999/xhtml">
            <p>CHARGE: 0</p>
            <p>Comment: renamed O2 into dioxygen</p>
            <p>Curator: lenov</p>
            <p>FORMULA: O2</p>
          </body>
        </notes>
        <annotation>
          <rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:bqbiol="http://biomodels.net/biology-qualifiers/">
            <rdf:Description rdf:about="#_8da2f742-2592-4760-987a-e76067d9c0cf">
              <bqbiol:is>
                <rdf:Bag>
                  <rdf:li rdf:resource="http://identifiers.org/chebi/CHEBI:15379" />
                </rdf:Bag>
              </bqbiol:is>
              <bqbiol:is>
                <rdf:Bag>
                  <rdf:li rdf:resource="http://identifiers.org/kegg.compound/C00007" />
                </rdf:Bag>
              </bqbiol:is>
            </rdf:Description>
          </rdf:RDF>
        </annotation>
      </species>
      <species boundaryCondition="false" compartment="m" constant="false" fbc:charge="0" fbc:chemicalFormula="O2" hasOnlySubstanceUnits="false" id="M_o2_m" initialConcentration="0" metaid="_29410bae-c38c-41c2-83ba-ecb943c90f3b" name="dioxygen">
        <notes>
          <body xmlns="http://www.w3.org/1999/xhtml">
            <p>CHARGE: 0</p>
            <p>Comment: renamed O2 into dioxygen</p>
            <p>Curator: lenov</p>
            <p>FORMULA: O2</p>
          </body>
        </notes>
        <annotation>
[...]

PR #403, opened 2018-04-04
Fix/find transport reactions

### LINKED ISSUE #390: Memote ought to check annotations in addition to formulae to determine transport reactions.
#### Problem description
Taken from a conversation on The SBML L3 Flux Balance Constraints package discussion list <sbml-flux@lists.sourceforge.net> discussing that the "Hill system [is] not sufficient for formulas".

Hello everyone,

I don’t mean to derail the conversation, but since our tool (memote) was brought up I should at least explain our reasoning wrt the following exchange:

>[...] But there is now at least one software using it to detect transport reactions (two species in different compartments with the same chemicalFormula).
- @lenov

>Errr.. Ok, but why would that need the extension, and the much more pressing question: Why is that software not using annotations to do this? I.e. same database-reference (Chebi, KEGG, MetaCyc)? The Chemical formula is a (imo) really questionable property to use here...
- @tpfau

We’re using the chemicalFormula field as a means to determine transport reactions, because we deemed it to be the only reliable field at the time of implementation. The annotation coverage of metabolites within a reconstruction may differ, just like the ID namespace or description that is used on metabolites. We have been assuming the worst case scenario, where Met A in compartment 1 may be annotated differently to Met A in compartment 2, which is quite possible when the reconstruction has been built by consolidation of two reconstructions with differing origin. The chemicalFormula, due to the mass-balance requirement, seemed to be the only field that we could consistently rely on. Thomas, you do make a valid point though, which we will consider for the ongoing development of our tool. The only drawback I see from checking annotations in addition to checking formulae is the increased runtime, but this is something we will discuss internally.

Thanks for sharing your thoughts!

siddC:
* [x] fix #390, #384
* [x] tests added/passed

Transport reactions were previously found using only the `formula` field of metabolite objects in reactions. This fix also catches transport reactions where the `formula` field may be absent or malformed as long as the metabolites in question have a proper `annotation` field.

### DIFF
--- a/tests/test_for_support/test_for_helpers.py
+++ b/tests/test_for_support/test_for_helpers.py
@@ -296,6 +296,29 @@ def energy_transfer_annotations(base):
 
 
 @register_with(MODEL_REGISTRY)
+def labeled_reaction(base):
+    """Provide a model with a labeled tranport reaction."""
+    a = cobra.Metabolite("a")
+    b = cobra.Metabolite("b")
+    rxn = cobra.Reaction("rxn")
+    rxn.annotation["SBO"] = "SBO:0000655"
+    rxn.add_metabolites({a: -1, b: 1})
+    base.add_reactions([rxn])
+    return base
+
+
+@register_with(MODEL_REGISTRY)
+def unlabeled_reaction(base):
+    """Provide a model with a labeled tranport reaction."""
+    a = cobra.Metabolite("a")
+    b = cobra.Metabolite("b")
+    rxn = cobra.Reaction("rxn")
+    rxn.add_metabolites({a: -1, b: 1})
+    base.add_reactions([rxn])
+    return base
+
+
+@register_with(MODEL_REGISTRY)
 def converting_reactions(base):
     """Provide a model with a couple of converting reaction."""
     a = cobra.Metabolite("atp_c", compartment="c")
@@ -476,7 +499,9 @@ def biomass_metabolite(base):
     ("energy_transfer_formulae", 0),
     ("energy_transfer_annotations", 0),
     ("phosphotransferase_system_formulae", 1),
-    ("phosphotransferase_system_annotations", 0)
+    ("phosphotransferase_system_annotations", 0),
+    ("labeled_reaction", 1),
+    ("unlabeled_reaction", 0)
 ], indirect=["model"])
 def test_find_transport_reactions(model, num):
     """Expect amount of transporters to be identified correctly."""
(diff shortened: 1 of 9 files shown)


tests/test_for_support/test_for_helpers.py
```diff
@@ -63,7 +97,62 @@ def abc_pump(base):
...
+    adp.annotation["bigg.metabolite"] = "adp"
+    adp.annotation["biocyc"] = ["META:ADP", "META:CPD0-1651"]
+    adp.annotation["metanetx.chemical"] = "MNXM7"
+    adp.annotation["seed.compound"] = "cpd00008"
+
+    h.annotation["bigg.metabolite"] = "h"
+    h.annotation["biocyc"] = "META:PROTON"
+    h.annotation["metanetx.chemical"] = "MNXM1"
+    h.annotation["seed.compound"] = "cpd00067"
+
```
ChristianLieven: Same here, please simplify to only use one of each principal types of annotation (list vs single string)

tests/test_for_support/test_for_helpers.py
```diff
@@ -79,7 +168,56 @@ def proton_pump(base):
...
+    atp.annotation["bigg.metabolite"] = "atp"
+    atp.annotation["biocyc"] = ["META:ATP", "META:CPD0-1634"]
+    atp.annotation["metanetx.chemical"] = "MNXM13"
+    atp.annotation["seed.compound"] = "cpd00011"
+
+    adp.annotation["bigg.metabolite"] = "adp"
+    adp.annotation["biocyc"] = ["META:ADP", "META:CPD0-1651"]
+    adp.annotation["metanetx.chemical"] = "MNXM7"
+    adp.annotation["seed.compound"] = "cpd00008"
+
```
ChristianLieven: Here too, it is really not necessary to blow up the code like that. Let's keep it concise wherever possible.

tests/test_for_support/test_for_helpers.py
```diff
@@ -94,7 +232,41 @@ def phosphotransferase_system(base):
...
+    malt6p = cobra.Metabolite("malt6p_c", compartment="c")
+
+    pep.annotation["bigg.metabolite"] = "pep"
+    pep.annotation["biocyc"] = "META:PHOSPHO-ENOL-PYRUVATE"
+    pep.annotation["metanetx.chemical"] = "MNXM73"
+    pep.annotation["seed.compound"] = "cpd00061"
+
+    pyr.annotation["bigg.metabolite"] = "pyr"
+    pyr.annotation["biocyc"] = "META:PYRUVATE"
+    pyr.annotation["metanetx.chemical"] = "MNXM23"
```
ChristianLieven: And here too :)

ChristianLieven:
Will squash & merge it when it has passed all CI tests.

PR #406, opened 2018-04-12
Feat/matrix unit tests

### LINKED ISSUE #378: Fix the stoichiometric consistency tests
We accidentally fixed the nullspace function and it revealed a problem with the stoichiometric consistency methods by Gevorgyan et al.

Done: The methods should return the results expected from the paper.

### LINKED ISSUE #405: Add unit tests for ``matrix.py``
#### Problem description

Create new file `test_for_matrix.py` that runs tests to ensure matrix functions are working as intended. This will be useful when trying to fix #378.

siddC:
* [ ] fix #405
* [ ] tests added/passed

Adds unit tests for `matrix.py`. This will be useful when attempting to fix #378 in ensuring the nullspace match up as they should.

### DIFF
--- a/tests/test_for_support/test_for_matrix.py
+++ b/tests/test_for_support/test_for_matrix.py
@@ -15,7 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Ensure the expected functioning of ``memote.support.consistency``."""
+"""Ensure the expected functioning of ``memote.support.matrix``."""
 
 from __future__ import absolute_import
 
@@ -23,10 +23,8 @@ import cobra
 import pytest
 from cobra.exceptions import Infeasible
 
-import memote.support.consistency as consistency
 import memote.support.matrix as matrix
 from memote.utils import register_with
-import memote.support.consistency_helpers as con_helpers
 
 MODEL_REGISTRY = dict()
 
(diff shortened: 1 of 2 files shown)


tests/test_for_support/test_for_matrix.py
```diff
@@ -0,0 +1,197 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Ensure the expected functioning of ``memote.support.matrix``."""
+
+from __future__ import absolute_import
+
+import cobra
+import pytest
+from cobra.exceptions import Infeasible
```
ChristianLieven: Is this import used anywhere?

tests/test_for_support/test_for_matrix.py
```diff
@@ -0,0 +1,197 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Ensure the expected functioning of ``memote.support.matrix``."""
+
+from __future__ import absolute_import
+
+import cobra
```
ChristianLieven: No need to import all of cobra if you're only using `Metabolite` and `Reaction`.

tests/test_for_support/test_for_matrix.py
```diff
@@ -0,0 +1,197 @@
...
+from cobra.exceptions import Infeasible
+
+import memote.support.matrix as matrix
+from memote.utils import register_with
+
+MODEL_REGISTRY = dict()
+
+
+@register_with(MODEL_REGISTRY)
+def three_components_closed(base):
```
ChristianLieven: I don't quite understand why you make a distinction between 3 and 4 component models?

Considering they yield identical results in many of the tests below.

tests/test_for_support/test_for_matrix.py
siddC: No its not. Sorry I had copied the imports from another `test_for_` file and forgot to check if everything was necessary.

tests/test_for_support/test_for_matrix.py
siddC: The only difference between the two was that the 4-compartment system has an additional, irreversible reaction, while the 3-comp system only has reversible reactions. However given that the results don't really differ, it can be taken out if needed.

PR #411, opened 2018-04-17
Feat/test met production

### LINKED ISSUE #401: Test that all carbon containing metabolites can be produced in complete medium
This is the reverse check for #256 and should help in gap finding.

1. Open all exchanges.
2. For each metabolite add a demand and maximize its flux.
3. Check feasibility.

siddC:
* [x] fix #401 
* [x] tests added/passed

Add tests to test for artefacts/infeasible metabolites in models (i.e. metabolites that cannot be produced even with all model exchanges open).


siddC:
The Travis CI Build failed for py27 but passed everything else, and I'm not sure why

ChristianLieven:
I'll take a look at this ASAP @siddC, thanks for submitting another PR!

siddC:
Thanks for looking into the error! Also, I'm not sure if I did this right. It seemed like a straightforward task so I went ahead and finished it but could you or @Midnighter make sure this actually does what the task is asking for? I'm just a bit worried since I followed the steps Moritz laid out in #401 without having the opportunity to discuss before I made the pull request.

memote/support/consistency.py
```diff
@@ -548,6 +548,36 @@ def find_metabolites_consumed_with_closed_bounds(model):
...
+    ----------
+    model : cobra.Model
+        The metabolic model under investigation.
+
+    """
+    feasible_mets = []
+    helpers.open_boundaries(model)
+    for met in model.metabolites:
+        # Add a demand and maximize flux
+        demand = Reaction("demand_for_{}".format(met.id), upper_bound=1000)
```
Midnighter: Use `model.add_boundary(met, type="demand")` here instead.

memote/support/consistency.py
```diff
@@ -548,6 +548,36 @@ def find_metabolites_consumed_with_closed_bounds(model):
     return mets_consumed
 
 
+def find_metabolite_production_feasibility(model):
+    """
+    Return number of metabolites produced in complete medium.
```
Midnighter: Wouldn't it be more interesting to know the metabolites that cannot be produced in complete medium? This has a large overlap with blocked reactions, deadend- and orphan metabolties, and isolated metabolites, of course.

memote/support/consistency.py
```diff
@@ -548,6 +548,36 @@ def find_metabolites_consumed_with_closed_bounds(model):
...
+    feasible_mets = []
+    helpers.open_boundaries(model)
+    for met in model.metabolites:
+        # Add a demand and maximize flux
+        demand = Reaction("demand_for_{}".format(met.id), upper_bound=1000)
+        demand.add_metabolites({met: -1})
+        model.add_reactions([demand])
+        # Check feasability
+        model.objective = demand.id
+        solution = model.optimize()
```
Midnighter: Depending on the choice of what the function should return (I'm writing it for compounds that are 'blocked' here):

```python
try:
    model.slim_optimize(error_value=None)
except OptimizationError:
    blocked.append(met)
```

and then remove the below condition.

memote/support/consistency.py
siddC: Actually, that is what is displayed on all the test, in the form of:
`ann["data"] = set(model.metabolites) - find_metabolite_production_feasibility(model)`

siddC: I can change the function so it returns `infeasible_mets` however

Midnighter: Whatever works better for you. Just seemed more straight forward this way around.

HISTORY.rst
```diff
@@ -23,6 +23,8 @@ Next Release
 * Add tests for detecting gene annotations (and verifying they are in
   MIRIAM style)
 * Add unit tests for ``matrix.py`` in file ``test_for_matrix.py``.
+* Add test to detect infeasible metabolites (i.e. metabolites are cannot
```
ChristianLieven: I think calling these metabolites `infeasible metabolites` makes it a little cryptic. I'd like you to stick with the established naming pattern, despite it being longer i.e.
`find_metabolites_produced_with_closed_bounds`
`find_metabolites_consumed_with_closed_bounds`
hence:
`find_metabolites_not_produced_with_open_bounds`
`find_metabolites_not_consumed_with_open_bounds`

memote/suite/tests/test_consistency.py
```diff
@@ -303,6 +303,28 @@ def test_find_metabolites_consumed_with_closed_bounds(read_only_model):
     assert len(ann["data"]) == 0, ann["message"]
 
 
+@annotate(title="fixme", type="count")
+def test_find_metabolite_production_infeasibility(read_only_model):
```
ChristianLieven: See my comment above on the naming 'convention'

memote/suite/tests/test_consistency.py
```diff
@@ -303,6 +303,28 @@ def test_find_metabolites_consumed_with_closed_bounds(read_only_model):
     assert len(ann["data"]) == 0, ann["message"]
 
 
+@annotate(title="fixme", type="count")
```
ChristianLieven: Is this a reminder for yourself or for us to fix it?

A simple descriptive title such as "Number of Metabolites That Cannot Be Produced On Complete Medium" is perfectly fine for now.

memote/support/consistency.py
```diff
@@ -548,6 +548,37 @@ def find_metabolites_consumed_with_closed_bounds(model):
     return mets_consumed
 
 
+def find_metabolite_production_infeasibility(model):
```
ChristianLieven: Did you have a look both at `find_metabolites_produced_with_closed_bounds` and `find_metabolites_consumed_with_closed_bounds`? They call `helpers.run_fba` to run the optimization. It makes sense to use modularity wherever it exists to avoid bugs and to capitalize on existing unit tests. In fact, I think you may have saved yourself some time copy/pasting the code from there, modifying it slightly afterwards.

tests/test_for_support/test_for_consistency.py
```diff
@@ -725,6 +725,15 @@ def test_find_metabolites_consumed_with_closed_bounds(model, num):
     assert len(consumed) == num
 
 
+@pytest.mark.parametrize("model, num", [
+    ("textbook", 0),
```
ChristianLieven: I cannot seem to find the definition of this `textbook` model anywhere, where is it defined?

In addition, it's smart to always test the two obvious cases (i.e. hypothesis and null-hypothesis). The function you're testing is supposed to find metabolites that cannot be produced when all exchanges are opened. So you ought to provide two cases, one where this is true i.e. there are metabolites that cannot be produced with open exchanges and one where this is false i.e. one where all metabolites can be produced.

siddC: "textbook" is defined in `conftest.py`. Its the `EcoliCore.xml.gz` file, but I've changed this now with better tests as asked

siddC:
py27 still fails unfortunately

Midnighter:
Right, I meant to update memote `setup.cfg`, `setup.py`, `tox.ini`, and `.travis.yml` for a while now but it looks like pip is forcing our hands :smiley: I'll get on this ASAP.

Midnighter:
The problem is actually with `importlib_resources`.

```bash
Missing build time requirements in pyproject.toml for importlib_resources
```

PR #420, opened 2018-05-07
Feat/cache support functions mki

### LINKED ISSUE #325: Cache support functions
Some of our support functions that try to identify particular reactions, for example, are used by various tests. In order to bring down the overall time for the test suite we should implement caching (`lru_cache`).

siddC:
* [x] fix #325 
* [x] tests added/passed

Some memote support functions that try to identify particular reactions, for example, are used by various tests. In order to bring down the overall time for the test, caching is now implemented (using `lru_cache`). Functions that are cached include: 
* `find_transport_reactions`
* `find_biomass_reactions`
* `find_demand_reactions`
* `find_sink_reactions`
* `find_exchange_reactions`
* `get_interface`
* `get_internals`
* `find_pure_metabolic_reactions`

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -28,6 +28,8 @@ Next Release
 * Add unit tests for ``matrix.py`` in file ``test_for_matrix.py``.
 * Add tests ``find_metabolites_not_produced_with_open_bounds`` and
   ``find_metabolites_not_consumed_with_open_bounds``
+* Add test ``find_duplicate_metabolites_in_compartments`` to detect duplicate
+  metabolites in identical compartments
 * Cache heavily used support functions in ``helpers.py`` and 
   ``consistency_helpers.py``
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -28,8 +28,6 @@ Next Release
 * Add unit tests for ``matrix.py`` in file ``test_for_matrix.py``.
 * Add tests ``find_metabolites_not_produced_with_open_bounds`` and
   ``find_metabolites_not_consumed_with_open_bounds``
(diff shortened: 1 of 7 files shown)


siddC:
I can cache run FBA and FVA as well. The reason I didn't was because some functions like those depend on other reaction/metabolite parameters, so I wasn't sure how useful caching would have been for them. The cache of 2 came because @Midnighter said that it works better with a cache size of 2, but isn't necessary. I can change it back to 1 if need be

ChristianLieven:
Nah that's okay then. Although I'd be interested to see how much of a performance improvement this actually gives us if FBA or FVA are not included.

I suppose it is tricky to cache those in a meaningful way. You're right about how the input parameters (i.e. model constraints) likely change. Would be good to know if there are any places at all in the code where we run either of them in a way that made caching possible.

ChristianLieven:
Before I merge this then, could you perhaps give caching FBA/ FVA / other more intensive operations another thought?

siddC:
Sure, I'll try to add those in and let you know how it goes.

siddC:
Added caching for `run_fba`. FVA is only used once in memote and isn't a function defined anywhere (like `run_fba`) but just used via import from cobrapy, so I've left it as is. Beyond these two, are there any other intensive operations I should cache? Because with this, all the most frequently used operations are officially cached with the exception of 2, `find_converting_reactions` and `find_interchange_biomass_reactions`.  The reason they aren't cached is because they caused issues with being unhashable. At least in the case for `find_interchange_biomass_reactions`, most of the functions it calls are already hashed, so I don't think this is too big an issue.

PR #421, opened 2018-05-07
Feat/check duplicate mets in comps

### LINKED ISSUE #241: Check for duplicate metabolites in the same compartment
If a model is consistently annotated with references that allow us to uniquely identify chemical entities (only full-length InChI strings as far as I know), we can use this to identify duplicate metabolites i.e. where different IDs are used for the same chemical entity.

siddC:
* [x] fix #241 
* [x] tests added/passed

Add tests that check for duplicate metabolites in the same compartment (i.e. 2 ATP metabolites in the cytosolic compartment). This check is performed using inchikey annotations.

### DIFF
--- a/memote/suite/tests/test_basic.py
+++ b/memote/suite/tests/test_basic.py
@@ -446,3 +446,25 @@ def test_find_unique_metabolites(read_only_model):
         total of {} ({:.2%}) unique metabolites in the model: {}""".format(
             len(ann["data"]), ann["metric"], truncate(ann["data"])))
     assert len(ann["data"]) < len(read_only_model.metabolites), ann["message"]
+
+
+@annotate(title="Number of Unique Metabolites", type="count")
+def test_find_duplicate_metabolites_in_compartments(read_only_model):
+    """
+    Expect there to be zero duplicate metabolites in the same compartments.
+
+    Metabolites may be transported into different compartments, and so models
+    may have metabolites such H2O or ATP in multiple compartments. However,
+    within a compartment, there should be no such duplicate metabolites. This
+    test therefore expects that every metabolite in any particular compartment
+    has unique inchikey values.
+    """
+    ann = test_find_duplicate_metabolites_in_compartments.annotation
+    ann["data"] = basic.find_duplicate_metabolites_in_compartments(
+        read_only_model)
+    ann["metric"] = len(ann["data"]) / len(read_only_model.metabolites)
+    ann["message"] = wrapper.fill(
+        """There are a total of {} ({:.2%}) metabolites in the model which
+        have duplicates in the same compartment: {}""".format(
+            len(ann["data"]), ann["metric"], truncate(ann["data"])))
+    assert len(ann["data"]) == 0, ann["message"]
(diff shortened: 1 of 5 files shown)


memote/suite/tests/test_basic.py
```diff
@@ -375,34 +375,6 @@ def test_find_constrained_transport_reactions(read_only_model):
                                                truncate(ann["data"])))
 
 
-@annotate(title="Fraction of Transport Reactions without GPR", type="percent")
```
ChristianLieven: Is there any specific reason why you've deleted this function and re-added it?

To get a cleaner diff and proportionate line change statistics, in the future, please just append new code instead of putting it in the centre of a module.

memote/suite/tests/test_basic.py
```diff
@@ -446,3 +418,53 @@ def test_find_unique_metabolites(read_only_model):
...
             len(ann["data"]), ann["metric"], truncate(ann["data"])))
     assert len(ann["data"]) < len(read_only_model.metabolites), ann["message"]
+
+
+@annotate(title="Number of Unique Metabolites", type="count")
+def test_find_duplicate_metabolites_in_compartments(read_only_model):
+    """
+    Expect there to be zero duplicate metabolites in the same compartments.
+
+    Metabolites may be transported into different compartments, and so models
```
ChristianLieven: The main reason for having this test is to clean up merged models or models from automated reconstruction pipelines as these are prone to having identical metabolites from different namespaces (hence different IDs).

Please upgrade the description accordingly.

memote/suite/tests/test_basic.py
```diff
@@ -446,3 +418,53 @@ def test_find_unique_metabolites(read_only_model):
         total of {} ({:.2%}) unique metabolites in the model: {}""".format(
             len(ann["data"]), ann["metric"], truncate(ann["data"])))
     assert len(ann["data"]) < len(read_only_model.metabolites), ann["message"]
+
+
+@annotate(title="Number of Unique Metabolites", type="count")
```
ChristianLieven: Remember to change the title to reflect the purpose/ output of the test!

memote/suite/tests/test_basic.py
```diff
@@ -446,3 +418,53 @@ def test_find_unique_metabolites(read_only_model):
...
+    Metabolites may be transported into different compartments, and so models
+    may have metabolites such H2O or ATP in multiple compartments. However,
+    within a compartment, there should be no such duplicate metabolites. This
+    test therefore expects that every metabolite in any particular compartment
+    has unique inchikey values.
+    """
+    ann = test_find_duplicate_metabolites_in_compartments.annotation
+    ann["data"] = basic.find_duplicate_metabolites_in_compartments(
+        read_only_model)
+    ann["metric"] = len(ann["data"]) / len(read_only_model.metabolites)
```
ChristianLieven: The content of `ann['data']` is a list of tuples as opposed to a list of metabolites hence dividing this by a list of metabolites doesn't seem very meaningful to me. I think in this case I wouldn't calculate the metric.

memote/suite/tests/test_basic.py
ChristianLieven: Oh and please remember to make place this test in the appropriate section in the test_config.yml!!

PR #433, opened 2018-05-16
Feat/check duplicate reactions

### LINKED ISSUE #240: Check for duplicate reactions
Find identical reactions in the same compartments but with different IDs.

siddC:
* [x] fix #240 
* [x] tests added/passed

Finds duplicate reactions in model (given they are annotated using BRENDA, KEGG, BiGG, etc.)

### DIFF
--- a/memote/support/basic.py
+++ b/memote/support/basic.py
@@ -295,8 +295,11 @@ def find_duplicate_reactions_in_compartments(model):
                 for db in RXN_DB if db in rxn.annotation]
     for a, b in combinations(ann_rxns, 2):
         for db in RXN_DB:
-            if a[1][db] == b[1][db]:
-                duplicates.append(a[0], b[0])
+            try:
+                if a[1][db] == b[1][db]:
+                    duplicates.append(a[0], b[0])
+            except KeyError:
+                continue
     return duplicates
 
 
--- a/memote/support/basic.py
+++ b/memote/support/basic.py
@@ -273,7 +273,7 @@ def find_duplicate_metabolites_in_compartments(model):
     return duplicates
 
 
-def find_duplicate_reactions_in_compartments(model):
+def find_duplicate_reactions(model):
     """
     Return list of reactions with duplicates.
 
--- a/memote/support/basic.py
+++ b/memote/support/basic.py
@@ -297,9 +297,9 @@ def find_duplicate_reactions(model):
         for key in rxn_db_identifiers:
             if key in rxn.annotation:
                 if type(rxn.annotation[key]) is list:
-                    ann_rxns.append([(
+                    ann_rxns.append((
                         rxn, frozenset((
-                            key, elem) for elem in rxn.annotation[key]))])
+                            key, elem) for elem in rxn.annotation[key])))
                 else:
                     ann_rxns.append((rxn,
                                      frozenset((key, rxn.annotation[key]))))
--- a/memote/support/basic.py
+++ b/memote/support/basic.py
@@ -298,7 +298,7 @@ def find_duplicate_reactions(model):
         for db in RXN_DB:
             try:
                 if a[1][db] == b[1][db]:
-                    duplicates.append(a[0], b[0])
+                    duplicates.append((a[0], b[0]))
             except KeyError:
                 continue
     return duplicates
--- a/memote/support/basic.py
+++ b/memote/support/basic.py
@@ -273,12 +273,6 @@ def find_duplicate_metabolites_in_compartments(model):
     return duplicates
 
 
-def check_transport_reaction_gpr_presence(model):
-    """Return the list of transport reactions that have no associated gpr."""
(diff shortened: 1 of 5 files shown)


siddC:
just rebased this branch too!

memote/support/basic.py
```diff
@@ -273,6 +273,36 @@ def find_duplicate_metabolites_in_compartments(model):
...
+    convert G6P(c) to F6P(c) in the same compartments.
+
+    Parameters
+    ----------
+    model : cobra.Model
+        The metabolic model under investigation
+
+    """
+    duplicates = []
+    RXN_DB = ["metanetx.reaction", "kegg.reaction", "brenda", "rhea", "biocyc",
```
Midnighter: All capital variable names are reserved for global variables. How about `rxn_db_identifiers`?

memote/support/basic.py
```diff
@@ -273,6 +273,36 @@ def find_duplicate_metabolites_in_compartments(model):
...
+              "bigg.reaction"]
+    ann_rxns = [(rxn, rxn.annotation) for rxn in model.reactions
```
Midnighter: This does a little duplicated work as your storing pairs of reactions and their complete annotation for every database. So if a reaction has annotations for four of those databases, for example, then you will have the same pair four times in that list.

Overall, I have two ideas to improve the below duplicate checking code:

1. Construct a list of pairs of the reaction and all its annotations in a set. I'm constructing a set of pairs of database identifier and reaction identifier here just in case two different reactions could have the same identifier in different databases. (This code is untested just off the top of my head.)

```python
ann_rxns = [(rxn, frozenset((db_id, rxn_id) for db_id in RXN_DB for rxn_id in rxn.annotation[db_id]) for rxn in model.reactions]

for (rxn_a, ann_a), (rxn_b, ann_b) in combinations(ann_rxns, 2):
    if len(ann_a & ann_b) > 0:
       duplicates.append((rxn_a, rxn_b))
```

2. The second idea is maybe for later as it involves creating RInChl keys for all the reactions. Such keys are unique based on the structure (InChl key) of the metabolites and the stoichiometry.

siddC:
We don't quite test for duplicate metabolites the same way. The only annotation that is used are inchikey values, so as long as the metabolites have unique inchikey values they are not appended. However, if they have the same inchikey values, they are both appended to the list of duplicates as a tuple element.

memote/support/basic.py
siddC: I like the second option more. That way, everything is depending on the structure and stoichiometry of individual metabolites in the model.

Midnighter: I totally agree but computing RInChl keys is not that easy and as far as @joaocardoso tells me the corresponding package to do so is a horrible C extension. Option 1 should be a quick improvement to the current code.

siddC: Okay yeah, I'll try that first. My alternative was to reaction a tuple `(rxn, dict)` where dict would be a dictionary with metabolite inchikey ids, and the values would be their stoichiometry in the reaction.

Midnighter: That also sounds good but we can't assume to have InChl keys for every metabolite in every model, can we?

memote/support/basic.py
```diff
@@ -291,15 +291,14 @@ def find_duplicate_reactions(model):
...
-    ann_rxns = [(rxn, rxn.annotation) for rxn in model.reactions
-                for db in rxn_db_identifiers if db in rxn.annotation]
-    for a, b in combinations(ann_rxns, 2):
-        for db in rxn_db_identifiers:
-            try:
-                if a[1][db] == b[1][db]:
-                    duplicates.append((a[0], b[0]))
-            except KeyError:
-                continue
+    ann_rxns = [(rxn, frozenset((key, rxn.annotation[key])
```
Midnighter: Good thinking adding the condition :smiley: The only thing missing now is how to handle the case when there are multiple identifiers for one database. I'm not 100% sure now, that definitely happens for metabolites but maybe not for reactions? If it does happen, you'd have to add another small loop.

memote/support/basic.py
```diff
@@ -273,6 +273,43 @@ def find_duplicate_metabolites_in_compartments(model):
...
+    for rxn in model.reactions:
+        for key in rxn_db_identifiers:
+            if key in rxn.annotation:
+                if type(rxn.annotation[key]) is list:
+                    tuple_list = [(rxn, frozenset((key, elem)))
+                                  for elem in rxn.annotation[key]]
+                    ann_rxns += tuple_list
+                else:
+                    ann_rxns.append((rxn,
+                                     frozenset((key, rxn.annotation[key]))))
```
Midnighter:
```python
ann_rxns.append((rxn, frozenset((key, elem) for elem in rxn.annotation[key])))
```

siddC:
Branch has just been rebased!

PR #437, opened 2018-05-25
Fix/yaml ids

### LINKED ISSUE #324: Store Identifiers in YAML or CSV
#### Problem description

>I think that a better approach would be to store the identifiers that are currently in global variables (as done for energy couples and in other places) in YAML or CSV files. This script could then load those files and generate the shortlist from there. Seems more future proof to me.
- @Midnighter on https://github.com/opencobra/memote/pull/321#pullrequestreview-90508859

siddC:
* [x] fix #324 
* [x] tests added/passed

Identifiers are now stored in YAML shortlist instead of a JSON shortlist for better readability


siddC:
@Midnighter the reason this build seems to fail is because the YAML file generated from running `generate_mnx_shortlists.py` (named `met_id_shortlist.yml`) does not appear to exist in the `tox` environment. Is there a place this file needs to be registered?

Midnighter:
Yes, in the `setup.cfg` line 83 onwards you need to list the export of YAML data:
```ini
[options.package_data]
memote.experimental.schemata = 
	*.json
memote.suite = 
	tests/*.py
memote.suite.templates = 
	*.html
	*.yml
memote.support.data = 
	*.csv
	*.json
```
Just add another line at the bottom with `*.yml`.

ChristianLieven:
> you need to list the export of YAML data

What is the idea behind this?

Midnighter:
By default Python will only package and install Python modules that it can find. So if you want other types of files to be part of your package that needs to be specified.

siddC:
Currently, the build passes for Python 2.7 but fails for 3.5 and 3.6. I'm not sure why that is the case

siddC:
Branch has just been rebased! Also, the codecov fails, but only because it misses a single line dealing with an `ImportError` exception, which I can't do anything about

siddC:
rebased the branch again as of yesterday. The codecov is still failing due to an `except ImportError` line in the code

Midnighter:
Cool, thank you. @ChristianLieven and I discussed a little more and we decided that a tabular format will be better in the long run. So I will probably build that on top of your PR unless you are motivated to re-work it (I can imagine that you're not :wink: ).

Midnighter:
I changed the PR around a bit. Modified the generating script and the way the shortlist is stored. Everything seems to run fine though all the energy generating cycles tests seem to be skipped. Was that always the case? Something that got broken by this? Can you take a look at that please @ChristianLieven?

Midnighter:
The latest pickle protocol is not Python 2&3 compatible. Thus I went back to JSON.

PR #444, opened 2018-06-04
Feat/scoring

### LINKED ISSUE #407: Swap colours such that green and a high number coincide
#### Problem description

Passing tests are green with 0% while failing tests are red but display 100% as the score. This should be inverted so that it is more sensible how the score is calculated.

ChristianLieven:
* [x] fix #351, fix #407

- Connects the score calculation with the report views. Both the history and the snapshot report will now display the total score. The snapshot report will also show subtotal scores per category in a bar chart.

- The colors for the tests have been inverted and the numbers are displayed as percent.

screen shot 2018-06-04 at 16 56 50


memote-report-app/src/app/report-data.service.ts
```diff
@@ -14,6 +14,7 @@ export class ReportDataService {
   scoredTests: string[] = [];
   reportType: string;
   allExpandState = false;
+  score: any;
```
Midnighter: Isn't score to be either `number` or `null`?

ChristianLieven:
> Are you actually doing any testing there?
@Midnighter

No, they are added automatically by `ng generate`. I'd have to look into that specifically to make it work, which could be worthwhile in the future!

memote-report-app/src/app/report-data.service.ts
ChristianLieven: No, `score` in this case is actually an object. It has the keys "sections" and "total_score" in a snapshot report, and "total_score" in the history report.

PR #448, opened 2018-06-13
Feat/diff

### LINKED ISSUE #346: Create a diff-report view
Create an Angular template that accepts two result files. It should show their performance side by side for easy comparison.

- [x] accept two results.json files
- [x] internally handle the results as results object files
- [x] show the results side-by-side
Optionally:
- [x] highlight only those results that differ

ChristianLieven:
* [x] fix #346 (issue number)

## Description
This PR creates an Angular template that accepts two result files. It shows their performance side by side for easy comparison. In theory, it could also accept more than 2 results, although I've not explicitly tested this.

The diff report:
accepts two results.json files
internally handles the results as results object files
show the results side-by-side
highlights only those results that differ


ChristianLieven:
There are likely still some bugs to be fixed and tests to be added which I'll do in this PR before merging!

ChristianLieven:
Oh and this it what it currently looks like:

screen shot 2018-06-13 at 09 38 57

memote/suite/cli/reports.py
```diff
@@ -153,11 +153,71 @@ def history(location, filename, custom_config):
...
+    pytest_args.extend(custom_tests)
+    config = ReportConfiguration.load()
+    # Update the default test configuration with custom ones (if any).
+    for custom in custom_config:
+        config.merge(ReportConfiguration.load(custom))
+    # Build the diff report specific data structure
+    diff_results = dict()
+    for model_path in models:
+        try:
+            model_filename = model_path.split('/')[-1].split('.')[0]
```
Midnighter: Use `os.path.basename` to get the last end of the path. This is not Windows compatible.

memote/suite/cli/reports.py
```diff
@@ -153,11 +153,71 @@ def history(location, filename, custom_config):
...
+            diff_results.setdefault(model_filename, dict())
+            model = callbacks._load_model(model_path)
+            model.solver = solver
+            _, diff_results[model_filename] = api.test_model(
+                model, results=True, pytest_args=pytest_args,
+                skip=skip, exclusive=exclusive, experimental=experimental)
+        except Exception as e:
```
Midnighter: This should be handled in the `callback._load_model` then. Let's find a way to aggregate and display those warnings.

memote/suite/reporting/diff.py
```diff
@@ -19,17 +19,81 @@
...
+    def __init__(self, diff_results, configuration, **kwargs):
         """Initialize the data."""
-        super(DiffReport, self).__init__(**kwargs)
+        super(DiffReport, self).__init__(
+            result=None, configuration=configuration, **kwargs)
+        self.config = configuration
         self._report_type = "diff"
+        self.result = self.format_and_score_diff_data(diff_results)
+        self.result.update(self.config)
+        print(jsonify(self.result, pretty=True))
```
Midnighter: Please remove this line when you're ready with the PR.

memote/suite/cli/reports.py
ChristianLieven: So should I just move the `try/except` statement over into `_load_model` or do we want to look at this in a separate PR maybe?

PR #471, opened 2018-07-12
refactor: optimize the history and report history

### LINKED ISSUE #466: Repair location for history and report
Done: I can run memote history or the report without changing git branches.

Midnighter:
* [x] fix #466 
* [x] tests added/passed

### DIFF
--- a/memote/suite/cli/reports.py
+++ b/memote/suite/cli/reports.py
@@ -109,10 +109,18 @@ def snapshot(model, filename, pytest_args, exclusive, skip, solver,
 
 @report.command(context_settings=CONTEXT_SETTINGS)
 @click.help_option("--help", "-h")
-@click.argument("location", envvar="MEMOTE_LOCATION")
+@click.option("--location", envvar="MEMOTE_LOCATION",
+              help="Location of test results. Can either by a directory or an "
+                   "rfc1738 compatible database URL.")
+@click.option("--model", envvar="MEMOTE_MODEL",
+              help="The path of the model file. Used to check if it was "
+                   "modified.")
 @click.option("--filename", type=click.Path(exists=False, writable=True),
               default="index.html", show_default=True,
               help="Path for the HTML report output.")
+@click.option("--deployment", default="gh-pages", show_default=True,
+              help="Results will be read from and committed to the given "
+                   "branch.")
 @click.option("--custom-config", type=click.Path(exists=True, dir_okay=False),
               multiple=True,
               help="A path to a report configuration file that will be merged "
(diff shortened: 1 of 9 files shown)


memote/suite/cli/reports.py
```diff
@@ -121,23 +129,21 @@ def snapshot(model, filename, pytest_args, exclusive, skip, solver,
...
-    to be found here. Or it can be an rfc1738 compatible database URL. Can
-    also be supplied via the environment variable
-    MEMOTE_DIRECTORY or configured in 'setup.cfg' or 'memote.ini'.
-
-    """
+def history(location, model, filename, deployment, custom_config):
+    """Generate a report over a model's git commit history."""
+    if model is None:
+        raise click.BadParameter("No 'model' path given or configured.")
+    if location is None:
```
ChristianLieven: Do I understand it correctly that this is not `None` when we run `memote report history` from within a git repo since it will pull the MEMOTE_LOCATION from the envvar?

memote/suite/cli/runner.py
```diff
@@ -83,7 +86,7 @@ def cli():
 @click.option("--filename", type=click.Path(exists=False, writable=True),
               default="result.json", show_default=True,
               help="Path for the collected results as JSON.")
-@click.option("--location", type=str, envvar="MEMOTE_LOCATION",
+@click.option("--location", envvar="MEMOTE_LOCATION",
```
ChristianLieven: Do I interpret this right, that users will still have to provide to do `--location Results` when they run any history related CLI commands?

I think it would be nice if we could leave out `--location` as an additional parameter when the user invokes the CLI from inside a memote git repo, but I'm OK with this too.

memote/suite/reporting/report.py
```diff
@@ -84,14 +84,8 @@ def render_html(self):
...
 
-    def determine_miscellaneous_tests(self):
-        """
-        Identify tests not explicitly configured in test organization.
-
-        List them as an additional card called `Misc`, which is where they will
-        now appear in the report.
-
-        """
+    def get_configured_tests(self):
```
ChristianLieven: I like this logical split!!

ChristianLieven:
I approve!! (forgot to do that explicitly on the review)

memote/suite/cli/reports.py
Midnighter: Also to your comment above, if the user has a `memote.ini` with, e.g.,
```ini
[memote]
location = Results
```
this will be picked up automatically and inserted as the parameter. `MEMOTE_LOCATION` in the environment works as well. So only if those are not found and the option is missing will the error be raised.

memote/suite/cli/runner.py
Midnighter: See above regarding configuration.

PR #475, opened 2018-07-17
refactor: improve secure travis var

Midnighter:
* Modify verbosity handling


Midnighter:
Not really more or less secure. Verbosity was just better command line argument handling that I snuck in here. It just handles diverse `.travis.yml` configurations better.

PR #486, opened 2018-08-30
Fix/various bugfixes

ChristianLieven:
This is a work in progress.

I'm hoping to collect a few more bugs before merging the code herein. I've just opened this PR to be able to rely on Travis CI!


HISTORY.rst
```diff
@@ -12,7 +24,6 @@ Next Release
 0.8.5 (2018-08-20)
 ------------------
 * Unpin cobra dependency and set it to >0.13.3.
-* Set ruamel.yaml to >=0.15 to keep up with cobra.
```
Midnighter: Did you rebase your branch on the latest devel? I'm confused why this line should be removed?

memote/suite/cli/reports.py
```diff
@@ -217,19 +226,29 @@ def diff(models, filename, pytest_args, exclusive, skip, solver,
...
+            loaded_models.append(model)
         except Exception as e:
             LOGGER.warning(
                 "The following exception occurred while loading model {}: {}"
                 "".format(model_filename, e))
+    # Running pytest in individual processes to avoid interference
+    partial_test_diff = partial(_test_diff, pytest_args=pytest_args,
+                                skip=skip, exclusive=exclusive,
+                                experimental=experimental)
+    pool = Pool(2)
```
Midnighter: Since you want to report any number of models >= 2, it'd be more appropriate to make the pool size
```python
min(len(models), os.cpu_count())
```

memote/suite/cli/reports.py
```diff
@@ -217,19 +226,29 @@ def diff(models, filename, pytest_args, exclusive, skip, solver,
...
         try:
             model_filename = os.path.basename(model_path)
             diff_results.setdefault(model_filename, dict())
             model = callbacks._load_model(model_path)
             model.solver = solver
-            _, diff_results[model_filename] = api.test_model(
-                model, results=True, pytest_args=pytest_args,
-                skip=skip, exclusive=exclusive, experimental=experimental)
+            loaded_models.append(model)
         except Exception as e:
```
Midnighter: Rewrite this catch statement to:
```python
except (IOError, SBMLError):
    LOGGER.debug(exc_info=True)
    LOGGER.warning("An error occurred while loading the model '%s'.", model_filename)
```
And abort the process if `len(loaded_models) < 2`.

Midnighter: Hopefully, all errors would be children of `IOError` but I'm not sure.

memote/suite/cli/reports.py
```diff
@@ -217,19 +226,29 @@ def diff(models, filename, pytest_args, exclusive, skip, solver,
...
                 "The following exception occurred while loading model {}: {}"
                 "".format(model_filename, e))
+    # Running pytest in individual processes to avoid interference
+    partial_test_diff = partial(_test_diff, pytest_args=pytest_args,
+                                skip=skip, exclusive=exclusive,
+                                experimental=experimental)
+    pool = Pool(2)
+    results = pool.map(partial_test_diff, loaded_models)
+
+    for model_path, result in zip(models, results):
```
Midnighter: Can you really rely on the order of `models` and `results` being the same? Somehow I doubt that.

memote/suite/results/result.py
```diff
@@ -38,3 +41,13 @@ def __init__(self, *args, **kwargs):
         super(MemoteResult, self).__init__(*args, **kwargs)
         self.meta = self.setdefault("meta", dict())
         self.cases = self.setdefault("tests", dict())
+        self.add_environment_information(self.meta)
```
Midnighter: I remember now why I didn't have this in the `__init__` method, when we are loading an existing result from file or database, we should not overwrite the environment information.

ChristianLieven: Gotcha! I've reorganised it so that the staticmethod is called explicitly only the when collecting results in the ResultCollectionPlugin which is instantiated only in `test_model`

memote/suite/cli/reports.py
```diff
@@ -234,15 +235,22 @@ def diff(models, filename, pytest_args, exclusive, skip, solver,
...
+        except (IOError, SBMLError):
+            LOGGER.debug(exc_info=True)
+            LOGGER.warning("An error occurred while loading the model '%s'. "
+                           "Skipping.", model_filename)
+    # Abort the diff report unless at least two models can be loaded
+    # successfully.
+    if len(loaded_models) < 2:
+        LOGGER.critical(
+            "Out of the {} provided models {} could be loaded. Check if the"
+            "models that could not e loaded are valid SBML. Aborting.")
```
Midnighter: Missing some arguments here, should be:
```python
        LOGGER.critical(
            "Out of the %d provided models only %d could be loaded. Please, check if the"
            "models that could not be loaded are valid SBML. Aborting.", len(models), len(loaded_models))
```

memote/suite/collect.py
```diff
@@ -66,7 +66,8 @@ def __init__(self, model, experimental_config=None,
         super(ResultCollectionPlugin, self).__init__(**kwargs)
         self._model = model
         self._exp_config = experimental_config
-        self.results = MemoteResult()
+        self.results = results = MemoteResult()
```
Midnighter: Not sure why you introduce the variable `results` in addition to `self.results`.

Midnighter:
Just two minor things then it's good to go!

memote/suite/cli/reports.py
ChristianLieven: Cheers! Forgot about that!

memote/suite/collect.py
ChristianLieven: Dunno :D I must have had a plan at the time. Maybe it was just a bit shorter. I've changed it back!

PR #508, opened 2018-10-08
Ignore commits in which the model was not changed with `memote run`, add option to trigger a test.

### LINKED ISSUE #500: feat: skip memote run in CI if .xml does not change
#### Problem description

As a user of memote as CI in my repo, it would be convenient if `memote run` would avoid doing all tests   for a commit in which the `.xml` file did not changed (e.g. added some functions, documentation, etc), something that is already included in `memote history`.

ChristianLieven:
* [x] fix #500 
* [x] description: Set ``memote run`` to default to skipping commits where the model was not 
  changed, but implement a flag that enables testing nonetheless. The flag is 
  ``--test-unchanged``.
* [x] tests added/passed: `test_run_test_unchanged_false`, `test_run_test_unchanged_true`

### DIFF
--- a/memote/support/basic.py
+++ b/memote/support/basic.py
@@ -24,6 +24,7 @@ from itertools import combinations
 from pylru import lrudecorator
 
 import memote.support.helpers as helpers
+from memote.utils import filter_none
 
 __all__ = ("check_metabolites_formula_presence",)
 
@@ -135,7 +136,7 @@ def find_ngam(model):
                  'ngam', 'non-growth', 'associated']
 
     refined_candidates = [rxn for rxn in candidates if any(
-        string in rxn.name.lower() for string in buzzwords
+        string in filter_none(rxn.name, '').lower() for string in buzzwords
     )]
 
     if refined_candidates:
--- a/memote/support/helpers.py
+++ b/memote/support/helpers.py
@@ -318,9 +318,8 @@ def find_biomass_reaction(model):
 
     biomass_met = []
     for met in model.metabolites:
-        if met.id.lower().startswith('biomass') or met.name.lower().startswith(
-            'biomass'
-        ):
+        if met.id.lower().startswith('biomass') \
+          or utils.filter_none(met.name, '').lower().startswith('biomass'):
             biomass_met.append(met)
     if biomass_met == 1:
         biomass_met_matches = set(
--- a/memote/support/helpers.py
+++ b/memote/support/helpers.py
@@ -318,8 +318,8 @@ def find_biomass_reaction(model):
 
     biomass_met = []
     for met in model.metabolites:
-        if met.id.lower().startswith('biomass') \
-          or utils.filter_none(met.name, '').lower().startswith('biomass'):
+        if 'biomass' in met.id.lower() or (
+                met.name is not None and 'biomass' in met.name.lower()):
             biomass_met.append(met)
     if biomass_met == 1:
         biomass_met_matches = set(
--- a/memote/utils.py
+++ b/memote/utils.py
@@ -22,7 +22,6 @@ from __future__ import absolute_import
 import json
 import logging
 from builtins import dict, str
-from os.path import basename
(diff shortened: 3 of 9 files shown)


memote/utils.py
```diff
@@ -138,6 +138,14 @@ def get_ids_and_bounds(iterable):
             elem in iterable]
 
 
+def filter_none(attribute, default):
```
Midnighter: This function is superfluous. Use `getattr(object, "attribute", default)` instead.

memote/support/basic.py
```diff
@@ -135,7 +136,7 @@ def find_ngam(model):
                  'ngam', 'non-growth', 'associated']
 
     refined_candidates = [rxn for rxn in candidates if any(
-        string in rxn.name.lower() for string in buzzwords
+        string in filter_none(rxn.name, '').lower() for string in buzzwords
```
Midnighter: See below, use `getattr(rxn, "name", "").lower()` instead of `filter_none`.

tests/test_for_suite/test_for_cli/test_for_runner.py
```diff
@@ -52,11 +53,41 @@ def test_run_output(runner, model_file):
 
 
 @pytest.mark.skip(reason="TODO: Need to provide input somehow.")
-def test_run_output(runner, tmpdir):
+def test_run_output_TODO(runner, tmpdir):
```
Midnighter: Is this TODO for a future PR?

ChristianLieven: This test was left over from before. I merely added TODO to the test here so that my pycodestyle doesn't complain about the c/p `test_run_output` function name anymore.

memote/utils.py
ChristianLieven: If I remember correctly, we talked about this briefly. I couldn't just use `getattr` since it checks for the presence of the attribute on the object. Even though the attribute may be missing (because optional) in SBML, cobrapy will instantiate all objects including the attributes, but defaults to giving them `None` values. For instance:

xml```
<species boundaryCondition="false" compartment="c" constant="false" fbc:chemicalFormula="C5H9O8P" hasOnlySubstanceUnits="false" id="M_r5p_c" metaid="M_r5p_c" sboTerm="SBO:0000247">
</species>
```

seems to result in something along these lines in the loaded cobrapy model object
```
model.metabolites.r5p_c.name
>
```

Comment (Midnighter) [reply memote/utils.py]: Right, my bad.

File: memote/support/helpers.py [hunk 4]
```diff
@@ -316,9 +316,8 @@ def find_biomass_reaction(model):
 
     biomass_met = []
     for met in model.metabolites:
-        if met.id.lower().startswith('biomass') or met.name.lower().startswith(
-            'biomass'
-        ):
+        if met.id.lower().startswith('biomass') \
```
Midnighter: We can make this condition slightly more generic if that's useful.
```python
        if 'biomass' in met.id.lower() or (met.name is not None and 'biomass' in met.name.lower()):
```

ChristianLieven: Fair enough.

ChristianLieven:
> I'm wondering though whether we should really make skipping commits the default behavior. I'm thinking about reversing the flag. It's easy for us to add the option to the .travis.yml but as a user I would expect memote run to always run the tests for me. Maybe @BenjaSanchez can provide some perspective.

I understand where you're coming from. It's not a problem to change it, and I'd like to do it now to merge my code tonight. @BenjaSanchez should you have strong feelings about it, we can certainly discuss some compromise, although for you implementing this should be as simple as changing the `travis.yml`.

BenjaSanchez:
@ChristianLieven absolutely fine with however the flag is defined, and I agree with @Midnighter that makes more sense to make the default to run anyways. As soon as this is in a new release I will adapt https://github.com/SysBioChalmers/yeast-GEM/pull/162 accordingly :)

PR #510, opened 2018-10-08
Identify specific MNX IDs (required for some tests) in the model by looking for matching annotations.

### LINKED ISSUE #485: memote could not find Metanetx ids in the annotation
Metabolite Ids are mostly in metacyc with c0 or e0 compartment suffix (e.g.:PPI_c0) , but metanetx ids are in the annotation field. When using essential_precursors_not_in_biomass function it could not find the essential precursors because of the suffixes, but neither could it find in the annotation field.

ChristianLieven:
* [X] fix #485 
* [X] tests added/passed


memote/support/helpers.py
```diff
@@ -693,28 +693,50 @@ def find_met_in_model(model, mnx_id, compartment_id=None):
...
-
     candidates = []
-    regex = re.compile('^{}(_[a-zA-Z]+?)*?$'.format(mnx_id))
+    # The MNX ID used in the model may or may not be tagged with a compartment
+    # tag e.g. `MNXM23141_c` vs. `MNXM23141`, which is tested with the
+    # following regex.
+    # If the MNX ID itself cannot be found as an ID, we try all other
+    # identifiers that are provided by our shortlist of MetaNetX' mapping
+    # table.
+    regex = re.compile('^{}(_[a-zA-Z0-9]+?)*?$'.format(mnx_id))
```
Midnighter: Since you're including the end of string mark `$` non-greedy matching should not be necessary. Also, we do expect at most one compartment suffix, right? My suggestion (untested)
`^{}(_[a-zA-Z0-9]+)?$` which should match exactly 0 or 1 underscore suffix.

memote/support/helpers.py
```diff
@@ -693,28 +693,50 @@ def find_met_in_model(model, mnx_id, compartment_id=None):
...
+    elif model.metabolites.query(compare_annotation, attribute='annotation'):
+        candidates = model.metabolites.query(
+            compare_annotation, attribute='annotation'
+        )
     else:
         for value in METANETX_SHORTLIST[mnx_id]:
             if value:
                 for ident in value:
-                    regex = re.compile('^{}(_[a-zA-Z]+?)*?$'.format(ident))
+                    regex = re.compile('^{}(_[a-zA-Z0-9]+?)*?$'.format(ident))
```
Midnighter: See above.

memote/utils.py
```diff
@@ -236,3 +236,18 @@ def jsonify(obj, pretty=False):
     except TypeError as error:
         log_json_incompatible_types(obj)
         raise_with_traceback(error)
+
+
+def flatten(list_of_lists):
```
Midnighter: I think we might change the annotation object to always contain lists in future.

memote/support/helpers.py
ChristianLieven: Done! Fingers crossed that all the tests still pass!

memote/utils.py
ChristianLieven: Sounds good. Until we do we should just go with this workaround.

PR #527, opened 2018-11-16
Test thermodynamics

### LINKED ISSUE #633: Create a 'slow' pytest mark and only run slow tests on Travis rather than locally in development.

ChristianLieven:
* [X] fix #633
* [X] tests added/passed

### DIFF
--- a/memote/suite/tests/test_thermodynamics.py
+++ b/memote/suite/tests/test_thermodynamics.py
@@ -27,15 +27,15 @@ if version_info[:2] < (3, 5):
     pytest.skip("Thermodynamic tests require at least Python version 3.5.",
                 allow_module_level=True)
 
-import memote.support.thermodynamics as thermo
-import memote.support.basic as basic
-from memote.utils import (annotate, get_ids, truncate)
+import memote.support.thermodynamics as thermo # noqa
+import memote.support.basic as basic # noqa
+from memote.utils import (annotate, get_ids, truncate) # noqa
 
 
 @annotate(title="Thermodynamic Reversibility of Purely Metabolic Reactions",
           format_type="percent")
 def test_find_incorrect_thermodynamic_reversibility(read_only_model):
-    """
+    u"""
     Expect reversibility of metabolic reactions to agree with thermodynamics.
 
     If a reaction is neither a transport reaction, a biomass reaction nor a
--- a/memote/suite/tests/test_thermodynamics.py
+++ b/memote/suite/tests/test_thermodynamics.py
@@ -43,13 +43,13 @@ def test_find_incorrect_thermodynamic_reversibility(read_only_model):
     This test checks if the reversibility attribute of each reaction
     in a list of cobrapy reactions agrees with a thermodynamics-based
     calculation of reversibility. To determine reversibility we calculate
-    the reversibility index lngamma of each reaction
-    using the eQuilibrator API. The default cutoff for lngamma of 3
+    the reversibility index ln_gamma of each reaction
+    using the eQuilibrator API. The default cutoff for ln_gamma of 3
     "corresponds to allowing concentrations to span three orders of magnitude
-    around 100 μM (~3 μM—3mM)" at "pH = 7, I = 0.1M and T = 298K". This means
-    that a reaction is considered irreversible if the concentration of an
-    individual metabolite would have to change more than three orders of
-    magnitude i.e. from 3µM to 3mM to reverse the direction of flux. For
+    around 100 μM (~3 μM—3 mM)" at "pH = 7, I = 0.1 M and T = 298 K". This
+    means that a reaction is considered irreversible if the concentration of
+    an individual metabolite would have to change more than three orders of
+    magnitude i.e. from 3 µM to 3 mM to reverse the direction of flux. For
     further information on the thermodynamic and implementational details
     please refer to
     https://doi.org/10.1093/bioinformatics/bts317 and
@@ -63,10 +63,11 @@ def test_find_incorrect_thermodynamic_reversibility(read_only_model):
     chemical groups which is essential for the calculation of Gibbs energy
     using component contributions. This test collects each exceptional reaction
     and returns them as a tuple containing each list in the following order:
-        1. Reactions with incorrect reversibility
-        2. Reactions with incomplete mapping to KEGG
-        3. Reactions with Metabolites that are problematic during calculation
-        4. Chemically or redox unbalanced Reactions (after mapping to KEGG)
+        1. Reactions with incorrect **rev**ersibility
+        2. Reactions with incomplete **mapping** to KEGG
+        3. Reactions with Metabolites that are problematic
+           during **calc**ulation
+        4. Chemically or redox un**bal**anced Reactions (after mapping to KEGG)
 
     This test simply reports the number of metabolic reactions that disagree
     with thermodynamic calculations i.e. are irreversible even though they
@@ -76,11 +77,11 @@ def test_find_incorrect_thermodynamic_reversibility(read_only_model):
     """
     ann = test_find_incorrect_thermodynamic_reversibility.annotation
     met_rxns = basic.find_pure_metabolic_reactions(read_only_model)
-    rev, map, calc, bal = \
+    rev, mapping, calc, bal = \
         thermo.find_incorrect_thermodynamic_reversibility(met_rxns)
-    ann["data"] = (get_ids(rev), get_ids(map), get_ids(calc), get_ids(bal))
+    ann["data"] = (get_ids(rev), get_ids(mapping), get_ids(calc), get_ids(bal))
     ann["message"] = "Out of {} purely metabolic reactions the reversibility "\
-                     "of {} does not agree with the calculated lngamma " \
+                     "of {} does not agree with the calculated ln_gamma " \
                      "cutoff ({:.2%}), and thus ought to be inverted. " \
                      "{} reactions " \
                      "could not be mapped to KEGG completely, " \
--- a/memote/suite/tests/test_thermodynamics.py
+++ b/memote/suite/tests/test_thermodynamics.py
@@ -19,6 +19,14 @@
 
 from __future__ import absolute_import, division
 
+from sys import version_info
+
+import pytest
+
+if version_info[:2] < (3, 5):
+    pytest.skip("Thermodynamic tests require at least Python version 3.5.",
+                allow_module_level=True)
+
 import memote.support.thermodynamics as thermo
 import memote.support.basic as basic
 from memote.utils import (annotate, get_ids, truncate)
--- a/memote/suite/tests/test_thermodynamics.py
+++ b/memote/suite/tests/test_thermodynamics.py
@@ -21,8 +21,7 @@ from __future__ import absolute_import, division
 
 import memote.support.thermodynamics as thermo
 import memote.support.basic as basic
-from memote.utils import (
-    annotate, get_ids, get_ids_and_bounds, truncate, wrapper)
+from memote.utils import (annotate, get_ids, truncate)
 
 
 @annotate(title="Thermodynamic Reversibility of Purely Metabolic Reactions",
--- a/memote/suite/tests/test_thermodynamics.py
+++ b/memote/suite/tests/test_thermodynamics.py
@@ -1,0 +1,89 @@
+# -*- coding: utf-8 -*-
+
+# Copyright 2018 Novo Nordisk Foundation Center for Biosustainability,
+# Technical University of Denmark.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Perform tests using eQuilibrator API on an instance of ``cobra.Model``."""
+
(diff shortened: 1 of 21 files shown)


memote/suite/tests/test_thermodynamics.py
```diff
@@ -0,0 +1,96 @@
...
+    u"""
+    Expect reversibility of metabolic reactions to agree with thermodynamics.
+
+    If a reaction is neither a transport reaction, a biomass reaction nor a
+    boundary reaction, it is counted as a purely metabolic reaction.
+    This test checks if the reversibility attribute of each reaction
+    in a list of cobrapy reactions agrees with a thermodynamics-based
+    calculation of reversibility. To determine reversibility we calculate
+    the reversibility index lngamma of each reaction
+    using the eQuilibrator API. The default cutoff for lngamma of 3
```
Midnighter: Nice and clear explanation! It is a little 'mathy', though, so I suggest you take a look at the ['Notes' section 12](https://numpydoc.readthedocs.io/en/latest/format.html#sections) for inspiration with including tex. For example, it wasn't clear on first look that lngamma is the natural logarithm of gamma.

memote/suite/tests/test_thermodynamics.py
```diff
@@ -0,0 +1,96 @@
...
+    "corresponds to allowing concentrations to span three orders of magnitude
+    around 100 μM (~3 μM—3mM)" at "pH = 7, I = 0.1M and T = 298K". This means
```
Midnighter: Is it immediately obvious to everyone what _pH_, _I_, and _T_ stand for?

memote/suite/tests/test_thermodynamics.py
```diff
@@ -0,0 +1,96 @@
...
+    individual metabolite would have to change more than three orders of
+    magnitude i.e. from 3µM to 3mM to reverse the direction of flux. For
+    further information on the thermodynamic and implementational details
+    please refer to
+    https://doi.org/10.1093/bioinformatics/bts317 and
+    https://gitlab.com/elad.noor/equilibrator-api/tree/master.
+
+    Please note that currently eQuilibrator can only determine the
+    reversibility index for chemically and redox balanced reactions whose
+    metabolites can be mapped to KEGG compound IDs (e.g. C00001). In addition
```
Midnighter: InChIs coming soon:tm: :smiley:

memote/suite/tests/test_thermodynamics.py
```diff
@@ -0,0 +1,96 @@
...
+import memote.support.thermodynamics as thermo # noqa
+import memote.support.basic as basic # noqa
+from memote.utils import (annotate, get_ids, truncate) # noqa
+
+
+@annotate(title="Thermodynamic Reversibility of Purely Metabolic Reactions",
+          format_type="percent")
+def test_find_incorrect_thermodynamic_reversibility(read_only_model):
+    u"""
+    Expect reversibility of metabolic reactions to agree with thermodynamics.
```
Midnighter: What about irreversible reactions that under the assumed conditions are pointing in the wrong direction? I think we should have a test for that, too.

memote/suite/tests/test_thermodynamics.py
```diff
@@ -0,0 +1,96 @@
...
+
+    This test simply reports the number of metabolic reactions that disagree
+    with thermodynamic calculations i.e. are irreversible even though they
+    should not be (and vice versa), considering the above fluctuations of
+    metabolite concentrations. It does not have a mandatory 'pass' criterium.
+
+    """
+    ann = test_find_incorrect_thermodynamic_reversibility.annotation
+    met_rxns = basic.find_pure_metabolic_reactions(read_only_model)
+    rev, map, calc, bal = \
```
Midnighter: I didn't write the code, so these variable names appear a bit cryptic to me.

memote/support/thermodynamics.py
```diff
@@ -0,0 +1,168 @@
...
+
+from six import string_types
+
+
+if version_info[:2] >= (3, 5):
+    from equilibrator_api import Reaction, CompoundMatcher
+    COMPOUND_MATCHER = CompoundMatcher()
+
+
+def smallest_compound_ID(kegg_ann_list):
```
Midnighter:
```suggestion
def smallest_compound_id(kegg_ann_list):
```

memote/support/thermodynamics.py
```diff
@@ -0,0 +1,168 @@
...
+    KEGG IDs may map to compounds, drugs and glycans prefixed respectively
+    with "C", "D", and "G" followed by at least 5 digits.
+
+    Parameters
+    ----------
+    kegg_ann_list: list
+        A list of mixed KEGG IDs.
+
+    """
+    only_cIDs = [x for x in kegg_ann_list if "C" in x]
```
Midnighter:
```suggestion
    only_compound_ids = [x for x in kegg_ann_list if x.startswith("C")]
```

memote/support/thermodynamics.py
Midnighter: Please maintain pep8 standard.

memote/support/thermodynamics.py
```diff
@@ -0,0 +1,168 @@
...
+    only_cIDs.sort(key=lambda x: int(x.lstrip('C')))
```
Midnighter:
```suggestion
    only_compound_ids.sort(key=lambda x: int(x.lstrip('C')))
```
Could replace `x` with `c` to be slightly more semantic.

memote/support/thermodynamics.py
Midnighter: See below for one-liner.

memote/support/thermodynamics.py
Midnighter: See below for one-liner.

memote/support/thermodynamics.py
```diff
@@ -0,0 +1,168 @@
...
+    return only_cIDs
+
+
+def get_equilibrator_rxn_string(rxn):
```
Midnighter:
```suggestion
def get_equilibrator_reaction_string(reaction):
```
My personal preference: Keep external names, i.e., parameters, functions, classes and their methods, and global variables verbose. Use abbreviations inside scope hidden from users.

memote/support/thermodynamics.py
```diff
@@ -0,0 +1,168 @@
...
+    kegg_rxn = rxn.reaction
+    for met in rxn.metabolites:
+        kegg_ann_id = met.annotation.get("kegg.compound")
+        if isinstance(kegg_ann_id, string_types) and "C" in kegg_ann_id:
+            kegg_rxn = kegg_rxn.replace(met.id, kegg_ann_id)
+        elif type(kegg_ann_id) is list and any("C" in s for s in kegg_ann_id):
+            kegg_rxn = kegg_rxn.replace(
+                met.id, smallest_compound_ID(kegg_ann_id)[0]
+            )
+        elif not getattr(met, "name"):
```
Midnighter: If `met` is a `cobra.Metabolite` the `name` attribute will always exist although it might be `None`.
```suggestion
        elif not met.name:
```

memote/support/thermodynamics.py
```diff
@@ -0,0 +1,168 @@
...
+    in the original reaction string.
+
+    Parameters
+    ----------
+    rxn: cobra.Reaction
+        The metabolic reaction under investigation.
+
+    """
+    kegg_rxn = rxn.reaction
+    for met in rxn.metabolites:
```
Midnighter: I would structure this helper function differently. You are calling this function for each reaction `r` and looping through its metabolites `m[r]`. So that means you construct `O(r * m[r])` mappings. I would create a mapping object only from the purely metabolic metabolites and pass that into this function to construct the reaction from. That means you first loop through metabolites `m` after which you loop through reactions `r` and thus  `O(r + m)` complexity. I have also [opened an issue](https://gitlab.com/elad.noor/equilibrator-api/issues/18) to create some kind of adapter between `cobra.Reaction` and equilibrator reactions.

memote/support/thermodynamics.py
```diff
@@ -0,0 +1,168 @@
...
+                kegg_match_id = df['CID'].iloc[0]
+                kegg_rxn = kegg_rxn.replace(met.id, kegg_match_id)
+            except Exception:
+                pass
+    # COBRApy reaction strings seem to use slightly different arrows which
+    # are not recognized by the eQuilibrator-API
+    return kegg_rxn.replace('-->', '->').replace('<--', '<-')
+
+
+def find_incorrect_thermodynamic_reversibility(reactions, lngamma=3):
```
Midnighter:
```suggestion
def find_incorrect_thermodynamic_reversibility(reactions, ln_gamma=3):
```

memote/support/thermodynamics.py
```diff
@@ -0,0 +1,168 @@
...
+    u"""
+    Return reactions whose reversibilities do not agree with thermodynamics.
+
+    This function checks if the reversibility attribute of each reaction
+    in a list of cobrapy reactions agrees with a thermodynamics-based
+    calculation of the reversibility. To determine reversibility we calculate
+    the reversibility index lngamma (see [1]_ section 3.5) of each reaction
+    using the eQuilibrator API [2]_. The default cutoff for lngamma
+    "corresponds to allowing concentrations to span three orders of magnitude
+    around 100 μM (~3 μM—3mM)" at "pH = 7, I = 0.1M and T = 298K" (see [1]_
```
Midnighter:
```suggestion
    around 100 μM (~3 μM—3 mM)" at "pH = 7, I = 0.1 M and T = 298 K" (see [1]_
```
SI units should have a space between value and unit.

Midnighter: Probably also in above docstring.

memote/support/thermodynamics.py
```diff
@@ -0,0 +1,168 @@
...
+
+    for rxn in reactions:
+        try:
+            eq_rxn = Reaction.parse_formula(get_equilibrator_rxn_string(rxn))
+        except Exception:
+            incomplete_mapping.append(rxn)
+            continue
+        if eq_rxn.check_full_reaction_balancing():
+            try:
+                ln_RI = eq_rxn.reversibility_index()
```
Midnighter:
```suggestion
                ln_rev_idx = eq_rxn.reversibility_index()
```

memote/support/thermodynamics.py
```diff
@@ -0,0 +1,168 @@
...
+            except Exception:
+                problematic_calculation.append(rxn)
+                continue
+            if (ln_RI < lngamma) != rxn.reversibility:
```
Midnighter:
```suggestion
            if (ln_rev_idx < lngamma) != rxn.reversibility:
```

memote/suite/tests/test_thermodynamics.py
ChristianLieven: I think every student of natural sciences will have had the basics of physics and chemistry to understand pH and T (especially when the value is above 200 and is in kelvin), but I do concede that `I` is still cryptic to me. I'll add an explanation.

memote/suite/tests/test_thermodynamics.py
ChristianLieven: Would it motivate you if I put a link to your WIP repo there :P

memote/suite/tests/test_thermodynamics.py
ChristianLieven: You don't think it is enough to actually describe the output in that order in the docstring just above?

>1. Reactions with incorrect *rev*ersibility
  2. Reactions with incomplete *map*ping to KEGG
  3. Reactions with Metabolites that are problematic during *calc*ulation
  4. Chemically or redox un*bal*anced Reactions (after mapping to KEGG)

I'm happy to be more expressive but I also wanted to avoid spreading a single line over 3-4 lines just because of expressive variable names.

memote/support/thermodynamics.py
ChristianLieven: I'm a big fan of the adapter idea, but until then I'm adamant about keeping the crutch (as opposed to stalling this).

I'll consider your suggestion, shouldn't be too hard to do.

memote/support/thermodynamics.py
ChristianLieven: Yeah definitely also in the docstring. Likewise with the change to `lngamma`

memote/suite/tests/test_thermodynamics.py
ChristianLieven: Tex might actually look nice in the docs, but not so in the report which is where this string will end up in its raw form. Do you have any ideas how to still keep it readable?

I added an underscore for ln_gamma just like with the variable. That's the only part really requiring Tex right?

memote/suite/tests/test_thermodynamics.py
ChristianLieven: Frankly, I'm not entirely sure how to cover that with just the ln_gamma metric. I believe this would require some sort of combined check of dG and ln_gamma? As far as I understand ln_gamma only tells me if a reaction is, in theory, reversible but I'm not sure if it takes the specific direction into account, or just looks at the reactants and products kinda separately. 

In the interest of meeting my sprint goal I've created a new card for this and would like to defer any further discussion to when we plan that task.

Please see #530

memote/support/thermodynamics.py
ChristianLieven: The newest commit should have addressed your concerns here. I'm hence resolving this conversation.

memote/suite/tests/test_thermodynamics.py
ChristianLieven: Let me know if I can resolve this.

ChristianLieven: Definitely keeping that in mind! I think that will be great to show how the score is calculated too!

Midnighter:
I realize now that the best option could be build and maintain the metabolite mapping while performing `find_incorrect_thermodynamic_reversibility`. So you would have a check whether the metabolite already exists in the mapping and only create one otherwise. I think this is totally fine to go in, though.

PR #536, opened 2018-11-28
fix: rely on cobrapy's find_blocked_reactions

### LINKED ISSUE #518: Change the logic of the test for blocked reactions.
Depend on the COBRApy function.

Midnighter:
* [x] fix #518 
* [x] Rely on the cobrapy provided function rather than a memote defined one.
* [x] Tests removed as we rely on cobrapy correctness.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,7 @@ History
 
 Next Release
 ------------
+* Identify blocked reactions using the cobrapy built-in function.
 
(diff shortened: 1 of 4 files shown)


Midnighter:
It's actually available in cobrapy for a long time already. Just an embarrassing oversight on my part.

Yes, I need to become more atomic about my commits. Noted.

Issue #558, opened 2019-01-07
Include check for Smiles and Inchi annotation


ChristianLieven:
Our checks for metabolite annotation currently excludes InChI strings.
- [x] Inchi

PR #564, opened 2019-01-16
Front-end Improvements

### LINKED ISSUE #502: As a user I want to clearly see how the overall score is calculated in the report
#### Problem description

Right now it is not entirely clear what the weights are per section and how the subscores sum up into the main score. The front-end should clearly display this kind of information (and ideally reflect changes on the backend).

### LINKED ISSUE #526: Reduce prominence of single overall score on the memote reports.
#### Problem description
Please see my post here: https://github.com/opencobra/memote/issues/232#issuecomment-438300424

ChristianLieven:
* [x] fix #502, fix #526

Description:
This PR fixes a bunch of issues with the reports interface: The prominence of the score has been reduced and more explanation is given as to what the score is for. Moreover, the weights are now clearly indicated as multiplier 'badges' and sub-total calculations are shown. All of this should help to increase transparency and aid researchers to interpret the results. Furthermore, the 'Help' section has been renamed to 'Readme' and has expanded to include a few paragraphs on the diversity of SBML-encoded genome-scale metabolic networks (reconstructions, initialized models). The principal sections have been renamed and a brief explanation for each has been added.

screen shot 2019-01-16 at 10 55 24

screen shot 2019-01-16 at 10 55 52


rmtfleming:
Looks nice. Are you planning to distinguish between a reconstruction and a model? Also, try testing it with Recon3 and Recon3.0model to see what difference you get. They provide more exceptional cases. The VMH api should provide access to metabolite and reaction annotations. A test for the existence of an InChI for each metabolite would be good.

ChristianLieven:
@rmtfleming  Thank you for your feedback!

> Are you planning to distinguish between reconstruction and a model? 
For now, I'm not planning on going further than what we've done above by explaining the differences between reconstructions and initialized models in the Readme, explaining the two sides of the report and expanding the descriptions of each test in a separate PR. Altogether I think it should be pretty clear now that this difference exists and how to interpret the report with regard to either type.

> Also, try testing it with Recon3 and Recon3.0model to see what difference you get. They provide more exceptional cases. The VMH API should provide access to metabolite and reaction annotations. 
Do you have a link to an SBML version at hand? At least, for now, we want to avoid making comparatively slow API-calls for each of the model's components. However, others have mentioned this before so we are considering possibly allowing users to toggle this in the future.

> A test for the existence of an InChI for each metabolite would be good.
Absolutely, this is coming up in a separate PR #558

memote-report-app/src/app/test-result.model.ts
```diff
@@ -10,7 +10,7 @@ export class TestResult {
   public format_type: string;
 
   private errorFailsafe(result: string) {
-    if (this.result !== 'skipped' && !this.data ) {
+    if (this.result !== 'skipped' && (this.data == null || typeof this.data === 'undefined') ) {
```
Midnighter: Can `this.data` here actually be `'undefined'`? Also, did you use `this.data == null` instead of `this.data === null` on purpose?

rmtfleming:
https://www.vmh.life/#downloadview

PR #574, opened 2019-01-17
Fix disappearing

ChristianLieven:

Fixes an issue with the angular app ignoring parametrized tests that had an undefined/ empty `data` object. The test cases would disappear from the report view, but now show up as errored:

https://puu.sh/Cy9Zv/be2ec5c248.png


Midnighter:
Travis is confused and tries to find PR #573 which doesn't exist any longer.

PR #575, opened 2019-01-17
WIP: Fix differences

### LINKED ISSUE #572: introduce a threshold for consistent test outcomes
`find_metabolites_produced_with_closed_bounds` and `find_metabolites_consumed_with_closed_bounds` need to be changed to use a threshold rather than comparing to exactly zero. Otherwise they report different numbers on each run due to solver feasibility tolerance.

Midnighter:
* [x] fix #572
* [x] tests added/passed

Some tests were showing different results online and offline. This was due to numerical instabilities which should now be fixed.

### DIFF
--- a/memote/suite/tests/test_consistency.py
+++ b/memote/suite/tests/test_consistency.py
@@ -324,9 +324,9 @@ def test_find_reactions_unbounded_flux_default_condition(model):
     """
     # TODO: Arbitrary threshold right now! Update after meta study!
     ann = test_find_reactions_unbounded_flux_default_condition.annotation
-    unbounded_rxns, fraction, _ = \
+    unbounded_rxn_ids, fraction, _ = \
         consistency.find_reactions_with_unbounded_flux_default_condition(model)
-    ann["data"] = get_ids(unbounded_rxns)
+    ann["data"] = unbounded_rxn_ids
     ann["metric"] = fraction
     ann["message"] = wrapper.fill(
         """ A fraction of {:.2%} of the non-blocked reactions (in total {}
--- a/memote/suite/tests/test_consistency.py
+++ b/memote/suite/tests/test_consistency.py
@@ -63,7 +63,7 @@ def test_stoichiometric_consistency(read_only_model):
 @pytest.mark.parametrize("met", [x for x in consistency.ENERGY_COUPLES])
 @annotate(title="Erroneous Energy-generating Cycles", format_type="count",
           data=dict(), message=dict())
-def test_detect_energy_generating_cycles(read_only_model, met):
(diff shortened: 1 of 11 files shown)


memote/suite/tests/test_consistency.py
```diff
@@ -258,65 +256,67 @@ def test_find_disconnected(read_only_model):
...
+    # )
+    # ann["metric"] = len(ann["data"]) / len(model.metabolites)
+    # ann["message"] = wrapper.fill(
+    #     """A total of {} ({:.2%}) metabolites can be produced without the model
+    #     needing to consume any substrate: {}""".format(
+    #         len(ann["data"]), ann["metric"], truncate(ann["data"])))
+    # assert len(ann["data"]) == 0, ann["message"]
+
+
+# FIXME (Moritz Beber): These metabolites are a subset of the 'unconserved
```
ChristianLieven: Here, instead of commenting out the code I was thinking doing something along the lines of `skipping` the tests by default and removing them from the main interface. This would keep them accessible via the CLI.

memote/support/consistency.py
```diff
@@ -479,14 +484,17 @@ def find_disconnected(model):
     return [met for met in model.metabolites if len(met.reactions) == 0]
 
 
-def find_metabolites_produced_with_closed_bounds(model):
+def find_metabolites_produced_with_closed_bounds(model, threshold=1E-07):
```
ChristianLieven: Do you expect the thresholds to be different/specific for each test? Should we perhaps set it in a global variable?

memote/suite/tests/test_consistency.py
```diff
@@ -63,7 +63,7 @@ def test_stoichiometric_consistency(read_only_model):
 @pytest.mark.parametrize("met", [x for x in consistency.ENERGY_COUPLES])
 @annotate(title="Erroneous Energy-generating Cycles", format_type="count",
           data=dict(), message=dict())
-def test_detect_energy_generating_cycles(read_only_model, met):
+def test_detect_energy_generating_cycles(model, met):
```
ChristianLieven: Isn't a `model` a copy and thus more expensive than a `read_only_model`? Wouldn't that slow the tests down?

memote/suite/tests/test_consistency.py
Midnighter: Hardcoding a fix doesn't really make it more accessible. Other than that you see a message `SKIPPED` for the test.

Midnighter:
@ChristianLieven I think I addressed all your comments. Please take another look.

ChristianLieven:
Don't forget the entry to the history.rst!

PR #581, opened 2019-01-18
upgrade memote to be pytest 4.x compatible

### LINKED ISSUE #557: Fix deprecation of the pytest_namespace hook which now errors in pytest 4.1
Quick fix is to pin pytest to <4.1, the more long-term solution is to use `pytest_generates` hooks to parametrize during runtime.

### LINKED ISSUE #578: Unpin pytest
See the test failure in #574 we need to upgrade pytest in `tox.ini` and memote itself and make the necessary fixes. Also affects #557.

Midnighter:
* [x] fix #578, fix #557 
* [x] Internal changes
* [x] tests added/passed

### DIFF
--- a/memote/suite/tests/test_biomass.py
+++ b/memote/suite/tests/test_biomass.py
@@ -78,18 +78,27 @@ def test_biomass_consistency(read_only_model, reaction_id):
     defined to be equal to 1 g/mmol. Conforming to this is essential in order
     to be able to reliably calculate growth yields, to cross-compare models,
     and to obtain valid predictions when simulating microbial consortia. A
-    deviation by 1e-03 is accepted.
+    deviation from 1 - 1E-03 to 1 + 1E-06 is accepted.
     """
     ann = test_biomass_consistency.annotation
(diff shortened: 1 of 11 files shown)


memote/suite/tests/test_biomass.py
```diff
@@ -57,16 +56,16 @@ def test_biomass_presence():
...
+    ann["data"] = [
+        rxn.id for rxn in helpers.find_biomass_reaction(read_only_model)]
     ann["message"] = wrapper.fill(
         """In this model {} the following biomass reactions were
         identified: {}""".format(
             len(ann["data"]), truncate(ann["data"])))
     assert len(ann["data"]) > 0, ann["message"]
 
 
-# TODO: Mark this test as skipped if the biomass components lack formulas!
```
ChristianLieven: Did you open an issue for this now that you've removed it, or did you end up fixing this in one of the other PRs?

Midnighter: Okay, I have added an alternative message when this is the case. I decided against skipping.

PR #584, opened 2019-01-23
Test sbml version

### LINKED ISSUE #284: Handle different SBML versions
If feasible check for different supported versions in cobrapy or simply reject models.

### LINKED ISSUE #314: Test if the model is a valid SBML and cancel testing if not!!
#### Problem description
Model Validator should be run first => Test should check if the model is valid and only then proceed with memote testing!

Warnings etc from cobrapy could be caught and displayed in the terminal together with a warning that memote cannot reliably test model that does not to conform to the SBML standard format!

### LINKED ISSUE #554: Flag if model is outdated SBML version
- [ ] Parse version information from SBML file and flag if it is outdated in the report.

ChristianLieven:
* [x] fix #284, fix #554, fix #314 
* [x] tests added/passed

## Changelog:
* Read SBML files with modified parser that can collect the level, version and
  whether the FBC package is used.
* Validate the SBML structure with the libSBML python API if the parser errors
  and produce a simple SBML validation report.
* Add test cases that report on the level and version, and FBC availability
  through the memote reports.


ChristianLieven:
@Midnighter what was it that needed to be done here? Now that you've probably started the test suite again we can go through this in a classic review, right?

Midnighter:
I'm finishing up my work on this and I will try to put it into neat commits.

Issue #537, opened 2018-11-29
Additional rxn SBO terms that could skip mass balance tests


BenjaSanchez:
#### Problem description

As discussed in https://github.com/SysBioChalmers/yeast-GEM/issues/137, there are certain SBO terms that could be an indicator for memote to skip assessing the corresponding reactions in terms of mass-balance / stoichiometry consistency:

* **Encapsulating process:** parent of biomass pseudoreaction, _"An aggregation of interactions and entities into a single process"_. 
http://www.ebi.ac.uk/sbo/main/SBO:0000395

* **Pseudoreaction**: _"A conceptual process used for modeling purposes, often created solely to complete model structure, with respect to providing inflow or outflow of matter or material. **Unlike other reactions, pseudoreactions are not usually subjected to mass balance considerations.**"_
http://www.ebi.ac.uk/sbo/main/SBO:0000631

Would any of these two (or both) make sense to include in the skipping heuristic for reactions?

Midnighter:
* In my opinion SBO:0000631 is a perfect candidate for reactions that we should skip.
* To me, SBO:0000395 requires some clarification from the maintainers of the SBO. Both children of the term describe inherently unbalanced reactions, however, _"An aggregation of interactions and entities into a single process"_ does not directly imply imbalance. For example, an entire pathway condensed into one reaction should clearly be balanced. Disclaimer: I don't know the SBO well enough to say if there might be a better term for my example.

Issue #522, opened 2018-11-07
Use SBO to identify NGAM


sulheim:
#### Problem description
The title says it all: It would be nice if memote could recognize a reaction annotated with SBO:0000630 as the NGAM - reaction.
http://www.ebi.ac.uk/sbo/main/display?sboId=SBO:0000630

<details>
System Information
==================
OS         Windows
OS-release      10
Python       3.6.6

Package Versions
================
Jinja2                 2.10
click                   6.7
click-configfile      0.2.3
click-log             0.3.2
cobra                0.12.0
cookiecutter          1.6.0
depinfo               1.4.0
future               0.16.0
gitpython            2.1.11
goodtables            1.0.0
importlib-resources   1.0.1
memote                0.8.4
numpydoc              0.8.0
pandas               0.23.4
pip                    18.0
pygithub               1.40
pylru                 1.1.0
pytest                3.7.4
python-libsbml       5.17.0
requests             2.19.1
ruamel.yaml         0.14.12
setuptools           40.2.0
six                  1.11.0
sqlalchemy           1.2.11
sympy                   1.2
travis-encrypt        1.1.2
travispy              0.3.5
wheel                0.31.1
</details>

Issue #469, opened 2018-07-12
Improve SBO term handling


ChristianLieven:
#### Problem description

We're really specifically checking for a few node SBO terms and not yet for the presence of any children of those nodes. Ideally, we'd have some internal representation of the ontology such that we can check if the the SBO term of an object matches a specific node term or any of its children.

We should also remove the duplicate check for SBO term annotations in our annotations tests as per this issue #447 

- [ ] Refactor tests and helper functions such they make use of the full SBO Ontology
- [ ] Explicitly exclude SBO terms from tests using the annotation.

Midnighter:
Very much agreed. I'd like to see the SBO handled in cobrapy, though.

gregmedlock:
To get the discussion going about what this will look like for cobrapy and memote, how do you envision cobrapy users interacting with SBO terms? There are many cases that automatic assignment makes sense, e.g. all metabolites should get SBO:0000247 (simple chemical entity), but these still have exceptions that make automation dangerous (e.g. polymers shouldn't have SBO:0000247).

Would a set of SBO helper functions be appropriate, such as functions that flag model elements as candidates for SBO terms and then let the user add the terms manually? Or is this getting into the realm of a package dedicated to annotating cobrapy models with SBO terms?

On the memote side, it seems like cobrapy-based "checks" could clean things up, e.g. calling `cobra.annotation.sbo.check_element()` to see whether an SBO term is being used properly, to the extent that it can be automatically assessed. If the entirety of SBO was integrated within cobrapy, this could also be done at the time of assignment (e.g. raise a warning if a non-transport reaction is assigned to a transport-related SBO term). But, the checks within memote would still be necessary because other packages probably wouldn't implement the practice.

Midnighter:
I like where you're going in terms of functionality.

Making it easy for a model reconstruction pipeline to mostly do the right thing is definitely a worthy goal.

> Would a set of SBO helper functions be appropriate, such as functions that flag model elements as candidates for SBO terms and then let the user add the terms manually? Or is this getting into the realm of a package dedicated to annotating cobrapy models with SBO terms?

[...]

Issue #315, opened 2018-01-16
Verify annotation IDs!


ChristianLieven:
#### Problem description

Namespaces/ Identifiers may change over time/ become deprecated! Warn the user for this possibility => False positives!

> FYI, here is an example of a KEGG ID that was given in a previous model (iMK1208) but is now obsolete. By matching reaction IDs in MetaCyc we could still find the new KEGG compound ID. Of course Memote isn't expected to check these kind of things, but good to be aware of this when you write some documentation.
- Eduard Kerkhoven

![image](https://user-images.githubusercontent.com/8027271/35000854-b98cfcde-fae5-11e7-8cd4-cb0d6b213f60.png)

This would fall right through since both match the `kegg.compund` regex
A test that parses `identifiers.org` could verify the existance of each identifer. Could have it skipped by default, and users who know that they can spare the time can activate it specifically

gregmedlock:
Looks like the most efficient method to do this would be the /identifers/validate/{id} service described here: http://identifiers.org/restws

I'm not familiar enough with REST/associated internet protocols to know whether having users with models containing ~1,000s-10,000s of identifiers to validate might overwhelm their servers...

ChristianLieven:
I'm not too familiar with this either. Would it be a good idea to reach out to the maintainers of identifiers.org about the implications of hitting them with quite a lot of requests?

draeger:
An alternative solution could be to do only one online check, namely to compare the MIRIAM XML file's date at http://identifiers.org/download/ to one that MeMoTe could ship (or contain). If the online version is newer, download it. Afterwards, all identifier schemas can be validated against the local XML file's content. This may be faster and save a lot of resources or online requests.

PR #587, opened 2019-01-23
switch order of list check to account for multiple SBO entries

gregmedlock:
Another small change in annotation checking--the previous line would lead to missing SBO terms for any object with more than one SBO term (e.g. it would end up trying things like ```['SBO:0000176', 'SBO:0000185'] in ['SBO:0000185']```).


gregmedlock:
Related to this and my previous PR (#586), it looks like annotation/SBO could use some additional tests. I'm happy to help brainstorm a bit of what is needed and implement them as I continue to work through some of these edge cases.

ChristianLieven:
Hi Greg! What were you thinking of? In general, I'm all ears for ideas on additional tests, but I'd also like  to point to some existing issues that need some love:

SBO:
- #469
- #522
- #537

Annotation:
- #315
- #187
- #186

memote/support/sbo.py
```diff
@@ -67,4 +67,4 @@ def check_component_for_specific_sbo_term(items, term):
     return [elem for elem in items if
             elem.annotation is None or
             'sbo' not in elem.annotation or
-            elem.annotation['sbo'] not in term]
+            term not in elem.annotation['sbo']]
```
ChristianLieven: Unfortunately, I don't think this fixes the issue entirely: 
- `test_transport_reaction_specific_sbo_presence` checks for a list of possible SBO terms related to transport reactions i.e. `term` can be a list of strings.
- The tests don't seem to cover that specific case

Would you be OK with taking that on? I realize that this is more than you bargained for.

ChristianLieven: Really the whole SBO is a temporary stand-in for a solution handling the ontology as suggested in #469

gregmedlock:
re: the issue with `test_transport_reaction_specific_sbo_presence`, it looks like that specific case is covered if any of the SBO transport terms are in `elem.annotation['sbo']`. I've updated the PR to check whether the terms are a list and handle accordingly. Seems a bit brute force but it gets the job done.

If more complex checks arise, e.g. check that multiple SBO terms are _all_ in an element's annotation, I think another function is warranted (e.g. `support.sbo.check_component_for_multiple_sbo_terms`).

I'll add tests for the changes in this PR and continue the higher level discussion in the other issues (#469 should probably come first).

In my second commit I accidentally slipped in an update to fix the gene annotation compliance check. If OK, I'll keep it in this PR.

gregmedlock:
One more clarification... I was planning on adding tests for the gene annotation compliance checks, but wanted to make sure I wasn't missing them somewhere. My intuition is that they should be located in `test_for_annotation.py` but just haven't been implemented yet. If this is correct, please let me know and I'll add the gene-equivalent of the existing metabolite/reaction tests within that file first thing tomorrow. Thanks!

ChristianLieven:
> One more clarification... I was planning on adding tests for the gene annotation compliance checks, but wanted to make sure I wasn't missing them somewhere. My intuition is that they should be located in test_for_annotation.py but just haven't been implemented yet.

1. Supporting functions for the `memote tests`, or `memote checks` if you prefer, go in `memote.support`,
2. The `memote tests` go in `memote.suite.tests`,
3. And `unit tests` for the support functions go in `tests.test_for_support`.

If I understand you correct you are talking about `unit tests` and in that case you are right to put them in `test_for_annotation`.

tests/test_for_support/test_for_sbo.py
```diff
@@ -70,10 +81,18 @@ def test_find_components_without_sbo_terms(model, num, components):
     ("specific_sbo_term", 0, "metabolites"),
     ("specific_sbo_term", 1, "reactions"),
     ("specific_sbo_term", 1, "genes")
```
ChristianLieven: A missing comma here leads to this error you're seeing on Travis:

```
==================================== ERRORS ====================================
___________ ERROR collecting tests/test_for_support/test_for_sbo.py ____________
tests/test_for_support/test_for_sbo.py:84: in <module>
    ("multiple_sbo_terms", 0, "metabolites"),
E   TypeError: 'tuple' object is not callable
```

tests/test_for_support/test_for_sbo.py
```diff
@@ -70,10 +81,18 @@ def test_find_components_without_sbo_terms(model, num, components):
...
+    ("multiple_sbo_terms", 0, "metabolites"),
+    ("multiple_sbo_terms", 1, "reactions"),
+    ("multiple_sbo_terms", 1, "genes")
 ], indirect=["model"])
 def test_find_components_without_specific_sbo_term(model, num, components):
```
ChristianLieven: I think it may be cleaner to take the `term` variable up into the parameters and only have a single assert statement in the test.

gregmedlock:
I think we're good to go with the latest commit!

PR #611, opened 2019-02-13
fix: handle a case where demand reactions already exist in a given model

ChristianLieven:

### DIFF
--- a/memote/support/biomass.py
+++ b/memote/support/biomass.py
@@ -131,8 +131,20 @@ def find_blocked_biomass_precursors(reaction, model):
     blocked_precursors = list()
     for precursor in precursors:
         with model:
-            dm_rxn = model.add_boundary(precursor, type="demand")
-            model.objective = dm_rxn
+            try:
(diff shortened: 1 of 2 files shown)


ChristianLieven:
I'd forgotten where we were doing this! Made changes accordingly.

memote/support/biomass.py
```diff
@@ -131,16 +131,16 @@ def find_blocked_biomass_precursors(reaction, model):
...
-            model.objective = dm_rxn
-            try:
-                solution = model.optimize()
-                LOGGER.debug(
-                    "%s: demand flux is '%g' and solver status is '%s'",
-                    str(precursor), solution.objective_value, solution.status)
-                if solution.objective_value <= 0.0:
-                    blocked_precursors.append(precursor)
-            except Infeasible:
+            _, ub = helpers.find_bounds(model)
```
Midnighter: The upper bound is pretty arbitrary here so it doesn't have to be a model-specific median. However, if you really want to use it, please move this line before the loop.

PR #614, opened 2019-02-19
refactor: add metrics to all test cases

Midnighter:
(With very few exceptions.)

* [x] tests added/passed

### DIFF
--- a/memote/suite/tests/test_biomass.py
+++ b/memote/suite/tests/test_biomass.py
@@ -68,6 +68,7 @@ def test_biomass_presence(model):
     ann = test_biomass_presence.annotation
     ann["data"] = [
         rxn.id for rxn in helpers.find_biomass_reaction(model)]
+    ann["metric"] = 1.0 - float(len(ann["data"]) > 0)
     ann["message"] = wrapper.fill(
         """In this model {} the following biomass reactions were
         identified: {}""".format(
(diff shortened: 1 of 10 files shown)


memote/suite/tests/test_biomass.py
```diff
@@ -309,17 +312,19 @@ def test_gam_in_biomass(model, reaction_id):
     """
     ann = test_gam_in_biomass.annotation
     reaction = model.reactions.get_by_id(reaction_id)
-    ann["data"][reaction_id] = biomass.gam_in_biomass(model, reaction)
-    ann["metric"][reaction_id] = 1.0  # Placeholder value.
-    if ann["data"][reaction_id]:
+    test_outcome = biomass.gam_in_biomass(model, reaction)
+    ann["data"][reaction_id] = test_outcome
+    test_outcome = ann["data"][reaction_id]
+    ann["metric"] = 1.0 - float(test_outcome)
```
ChristianLieven: The metric here is not per reaction_id, which may lead to issues in the report. I assume you did this on purpose so that you'd get one comparable metric value per test but did you try generating a snapshot or diff report with this?

memote/suite/tests/test_biomass.py
```diff
@@ -346,8 +351,9 @@ def test_fast_growth_default(model, reaction_id):
 
     """
     ann = test_fast_growth_default.annotation
-    ann["data"][reaction_id] = helpers.run_fba(model, reaction_id) > 2.81
-    ann["metric"][reaction_id] = 1.0  # Placeholder value.
+    test_outcome = helpers.run_fba(model, reaction_id) > 2.81
+    ann["data"][reaction_id] = test_outcome
+    ann["metric"] = 1.0 - float(test_outcome)
```
ChristianLieven: Same here. Your metric would be overwritten with each reaction that is tested. The final metric won't be representative of the entire test but just of the final iteration or not?

memote/suite/tests/test_biomass.py
Midnighter: No, oversight on my part. Thank you for catching this.

PR #616, opened 2019-02-26
Cli tests

ChristianLieven:
* [X] fix https://app.zenhub.com/workspaces/caffeine-platform-59dd1cab2fbdf4155c45e6e9/issues/dd-decaf/scrum/654
* [X] tests added/passed

- Removes our dependecy on PyGithub and TravisPy by reimplementing all the necessary steps with `requests`
- Adds unit and integration tests for `runner.py` and `reports.py`, except for `memote online`.
- Debugs `memote online` - could only be tested manually, CI is impossible without a secure way of storing github credentials.
- Add logging statements to all CLI functions to facilitate future debugging.


tests/conftest.py
```diff
@@ -45,6 +45,16 @@ def small_file(request):
         return join(dirname(__file__), "data", "EcoliCore.xml.gz")
 
 
+@pytest.fixture(scope="session", params=["invalid"])
```
kvikshaug: Why does this fixture accept parameters - it seems the invalid model is the only use case here?

tests/test_for_suite/conftest.py
```diff
@@ -40,6 +40,14 @@ def model_file(small_file, tmpdir_factory):
 
 
 @pytest.fixture(scope="session")
+def invalid_file(invalid_model, tmpdir_factory):
```
kvikshaug: The roles of the `invalid_model` and `invalid_file` fixtures are a bit unclear to me. The first seem to return the path to the invalid model, while the latter returns the path to a tmp copy of it. Is the tmp copy needed because tests could make modifications to the file? I would maybe consider naming the fixtures a bit differently or adding docstrings to make their functions clear.

memote/suite/cli/runner.py
```diff
@@ -51,6 +47,15 @@
     ResultManager, RepoResultManager, SQLResultManager, HistoryManager)
 from memote.utils import is_modified, stdout_notifications
 
+import requests
+from requests.exceptions import HTTPError
+import tempfile
+
+try:
+    from urllib.parse import quote_plus
```
kvikshaug: Since you use `six` throughout memote, could consider using https://pythonhosted.org/six/#module-six.moves here.

memote/suite/cli/runner.py
```diff
@@ -393,88 +411,282 @@ def history(model, message, rewrite, solver, location, pytest_args, deployment,
...
-    while not t_user.sync():
-        sleep(0.1)
+    else:
+        LOGGER.info("Success!")
+        t_user = response.json()
+
+    # Synchronize a User's projects between GitHub and Travis
+    LOGGER.info("Synchronizing user projects between GitHub and Travis.")
+    synced = False
+    while not synced:
```
kvikshaug: This is a bit odd, why do we need to retry indefinitely?

tests/test_for_suite/test_for_cli/test_for_runner.py
```diff
@@ -93,3 +137,69 @@ def test_run_skip_unchanged_true(runner, mock_repo):
...
+    result = runner.invoke(cli, [
+        "new", "--directory", target_dir], input=user_responses)
+    assert result.exit_code == 0
+
+
+@pytest.mark.skip(reason="Currently fails with: fatal: could not read "
+                         "Username for 'https://github.com': Device not "
+                         "configured. In addition, incorporating github "
+                         "credentials is a bit of an issue.")
+def test_online(runner, mock_repo, monkeypatch):
```
kvikshaug: If the idea is that this function should be tested manually from time to time, it could be helpful to document how to do that.

ChristianLieven:
> Really nice to see this increase in test coverage! I've added some minor comments inline just for consideration, but I also have two slightly more important concerns:
> 
> * The [`mock_repo`](https://github.com/opencobra/memote/blob/8b0c92ad789f01bda5df9b7c096f875e10bb4039/tests/test_for_suite/conftest.py#L50) fixture clones from your private account on GitHub. That adds a couple of seconds to each test requiring this repo, and assumes your mock repo will always exist. Would it be possible to clone it from test data, like with the models? (I realize this is perhaps outside of the scope of this PR, but I'd suggest maybe creating a new issue for it.)
> * The runtime of the test suite has gone up quite a bit. `test_new` and `test_history` alone adds 23 seconds. I would recommend to try to optimize the tests. If that's not possible, then consider flagging the very slow tests so that they're maybe not included in the local test suite by default, but are still run on CI. See the test duration summary below (note that the `setup` timings are mostly due to the `mock_repo` fixture):
> 
> ```
> 18.65s call     tests/test_for_suite/test_for_cli/test_for_runner.py::test_new
> 4.37s call     tests/test_for_suite/test_for_cli/test_for_runner.py::test_history
> 4.06s call     tests/test_for_suite/test_for_cli/test_for_runner.py::test_run_simple[ecoli-core]
> 3.79s call     tests/test_for_suite/test_for_cli/test_for_reports.py::test_diff[ecoli-core]
> 3.74s call     tests/test_for_suite/test_for_cli/test_for_runner.py::test_run_no_repo_ignore_git_false[ecoli-core]
> 3.68s call     tests/test_for_suite/test_for_cli/test_for_runner.py::test_run_output[ecoli-core]
> 2.94s call     tests/test_for_suite/test_for_cli/test_for_reports.py::test_snapshot[ecoli-core]
> 2.83s call     tests/test_for_suite/test_for_api.py::test_test_model_code[glpk-complete_failure-1]
> 2.83s call     tests/test_for_suite/test_for_cli/test_for_runner.py::test_run_skip_unchanged_false
> 2.79s setup    tests/test_for_suite/test_for_cli/test_for_runner.py::test_run_dirty_repo_ignore_git_false
> 2.66s call     tests/test_for_suite/test_for_api.py::test_test_model_result[glpk-complete_failure]
> 2.61s setup    tests/test_for_suite/test_for_cli/test_for_reports.py::test_history
> 2.61s call     tests/test_for_suite/test_for_api.py::test_snapshot_report_file[glpk-complete_failure]
> 1.85s setup    tests/test_for_suite/test_for_cli/test_for_runner.py::test_run_no_location
> 1.54s setup    tests/test_for_suite/test_for_cli/test_for_runner.py::test_history
> 1.44s setup    tests/test_for_suite/test_for_cli/test_for_runner.py::test_run_skip_unchanged_false
> 1.39s setup    tests/test_for_suite/test_for_cli/test_for_runner.py::test_run_skip_unchanged_true
> ```

To combat both, would I have to use `git submodules` then? And what about the drawbacks/ implications that has?

https://stackoverflow.com/questions/4659549/maintain-git-repo-inside-another-git-repo

kvikshaug:
> To combat both, would I have to use `git submodules` then? And what about the drawbacks/ implications that has?
> 
> https://stackoverflow.com/questions/4659549/maintain-git-repo-inside-another-git-repo

No, I wouldn't add it as a submodule. I think it's simplest to store the repo as a zip/tarball and in the fixture just uncompress it to a temporary folder.

memote/suite/cli/runner.py
ChristianLieven: The retry loop was part of the original implementation which I did not want to meddle with.

Should I terminate the loop after X iterations?

tests/test_for_suite/conftest.py
ChristianLieven: I frankly just replicated the implementation I already found, so as far as I understand this also applies to `small_file` and `model_file`. Perhaps, @Midnighter could shed some light as to why this is the way it is?

tests/conftest.py
ChristianLieven: I frankly just replicated the implementation I already found, so as far as I understand this also applies to `small_file` too. Perhaps, @Midnighter could shed some light as to why this is the way it is?

tests/test_for_suite/test_for_cli/test_for_runner.py
ChristianLieven: I'm hoping to be able to redefine 'origin' to point to another local repo. This way we could run this test routinely and this specific documentation may be superfluous.

tests/test_for_suite/conftest.py
ChristianLieven: > > To combat both, would I have to use `git submodules` then? And what about the drawbacks/ implications that has?
> > https://stackoverflow.com/questions/4659549/maintain-git-repo-inside-another-git-repo
> 
> No, I wouldn't add it as a submodule. I think it's simplest to store the repo as a zip/tarball and in the fixture just uncompress it to a temporary folder.

OK that seems reasonable.

ChristianLieven:
> > To combat both, would I have to use `git submodules` then? And what about the drawbacks/ implications that has?
> > https://stackoverflow.com/questions/4659549/maintain-git-repo-inside-another-git-repo
> 
> No, I wouldn't add it as a submodule. I think it's simplest to store the repo as a zip/tarball and in the fixture just uncompress it to a temporary folder.

OK, that seems reasonable.

memote/suite/cli/runner.py
kvikshaug: Ah.. that's hard to answer, so then I think we need to either dig into the original implementation to understand why this was required in the first place, or just leave it as it is.

Midnighter: Probably some number of max tries would be better. Maybe a total of 30 or 60 seconds is a good number.

tests/test_for_suite/test_for_cli/test_for_runner.py
```diff
@@ -93,3 +134,75 @@ def test_run_skip_unchanged_true(runner, mock_repo):
...
+                     "description\n2019-02-07\n2019\n0.1.0\n" \
+                     "default\ndefault\ngh-pages"
+    result = runner.invoke(cli, [
+        "new", "--directory", target_dir], input=user_responses)
+    assert result.exit_code == 0
+
+
+def test_online(runner, mock_repo, monkeypatch):
+    # Build-up
+    def mock_setup_gh_repo(github_repository, github_username, note):
```
Midnighter: You could use a catch all since you're not interested in the arguments.

```suggestion
    def mock_setup_gh_repo(*args, **kwargs):
```

tests/test_for_suite/test_for_cli/test_for_runner.py
```diff
@@ -93,3 +134,75 @@ def test_run_skip_unchanged_true(runner, mock_repo):
...
+    path2local = join(path2origin, pardir, 'cloned_repo')
+    os.mkdir(path2local)
+    localrepo = originrepo.clone(path2local)
+    os.chdir(path2local)
+
+    # Build context_settings
+    context_settings = ConfigFileProcessor.read_config()
+    github_repository = context_settings["github_repository"]
+    github_username = context_settings["github_username"]
+    print(github_username)
```
Midnighter: Seems a leftover from debugging.
```suggestion
```

tests/test_for_suite/test_for_cli/test_for_runner.py
Midnighter: I'm afraid you cannot have a situation where Travis will be able to connect to a local repo :( so I'm indeed also unsure when and how to run this test.

ChristianLieven: Any interaction with TravisCI (or GitHub) is being mocked in this test since we don't want to test API interactions. Only the final actions performed by `memote online` are essentially tested here:
- Loading the travis.yml
- Storing the token in the travis.yml
- Adding, committing and pushing to origin

Even though this isn't much I think it still warrants thorough test coverage.

memote/suite/cli/runner.py
```diff
@@ -393,88 +411,290 @@ def history(model, message, rewrite, solver, location, pytest_args, deployment,
...
-        when = user.created_at.isoformat(sep=" ")
-        LOGGER.info(
-            "Logged in to user '{}' created on '{}'.".format(user.login, when))
-    except BadCredentialsException:
+        LOGGER.info("Using your credentials to authenticate you on GitHub.")
+        response = requests.get(
+            'https://api.github.com/user', auth=credentials, headers=headers
+        )
+        response.raise_for_status()
+    except HTTPError:
```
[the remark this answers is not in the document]
ChristianLieven: > that should either be handled differently or not handled at all so that the command is aborted.

As far as I understand `raise_for_status()` raises an HTTPError for any response that isn't "200"? Is your issue with the logger output saying "incorrect username or password" even though it could also be any other "4xx/ 5xx" code or with the fact that I am not responding to each possible response code individually?

tests/test_for_suite/test_for_cli/test_for_runner.py
Midnighter: Oh, yes of course. That means you just extract the compressed repository to two different locations and declare one to be the origin for the other.

Maybe it's time we create a 'slow' pytest mark and only run slow tests on Travis rather than locally in development.

memote/suite/cli/runner.py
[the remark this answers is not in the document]
ChristianLieven: I think I got what you meant here. I'm now distinguishing between 401, other errors,and success.

memote/suite/cli/runner.py
```diff
@@ -393,88 +411,290 @@ def history(model, message, rewrite, solver, location, pytest_args, deployment,
...
+        response.raise_for_status()
         LOGGER.warning(
             "Using existing repository '{}'. This may override previous "
             "settings.".format(github_repository))
-    except UnknownObjectException:
-        gh_repo = user.create_repo(github_repository)
-    if note == "memote-ci access":
-        note = "{} to {}".format(note, github_repository)
+        gh_repo = response.json()
+    except HTTPError:
```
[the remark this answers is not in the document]
ChristianLieven: I'm now distinguishing between 404, other errors, and success.

memote/suite/cli/runner.py
```diff
@@ -393,88 +411,290 @@ def history(model, message, rewrite, solver, location, pytest_args, deployment,
...
+                "'{}' did not exist on GitHub yet."
+                " Creating it for you now!".format(github_repository))
+            response = requests.post(
+                'https://api.github.com/user/repos',
+                auth=credentials,
+                headers=headers,
+                json={"name": github_repository}
+            )
+            response.raise_for_status()
+        except HTTPError:
```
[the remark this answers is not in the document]
ChristianLieven: Done.

tests/test_for_suite/test_for_cli/test_for_reports.py
```diff
@@ -39,3 +41,29 @@ def test_snapshot(runner, model_file):
         "snapshot", "--filename", output, model_file])
     assert result.exit_code == 0
     assert exists(output)
+
+
+def test_diff(runner, model_file):
+    """Expect the diff report to function."""
+    output = model_file.split(".", 1)[0] + ".html"
```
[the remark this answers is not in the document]
ChristianLieven: Done.

tests/test_for_suite/test_for_cli/test_for_runner.py
```diff
@@ -93,3 +134,75 @@ def test_run_skip_unchanged_true(runner, mock_repo):
     number_of_result_files = len(os.listdir(join(mock_repo[0], 'results')))
     os.chdir(previous_wd)
     assert number_of_result_files == 3
+
+
+def test_new(runner, tmpdir):
+    """Expect memote new to create a cookiecutter repo."""
+    target_dir = str(tmpdir)
+    user_responses = "John\nj@d.com\nJD\nmock-repo\nmock-repo\n" \
```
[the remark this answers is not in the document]
ChristianLieven: Why is carrying out an I/O operation better than defining the variable in the code?

memote/suite/cli/runner.py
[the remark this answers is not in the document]
Midnighter: Sorry for being too short. Yes, I meant only displaying the message about incorrect username/password when the status code is 401 and otherwise the message of the error.

tests/test_for_suite/test_for_cli/test_for_runner.py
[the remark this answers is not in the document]
Midnighter: I was thinking in terms of ease of editability. However, you're right that additional IO should be avoided. So triple quotes are a good compromise I think.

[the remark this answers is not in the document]
ChristianLieven: Personally I'd rather explicitly state the newline characters than obscuring that information by using the triple quotes. In addition, I'd have to violate the indentation which makes it look more contentious to my eyes than the original single quote string.

I'm going to leave it as it is for now.

tests/test_for_suite/test_for_cli/test_for_runner.py
ChristianLieven: Done

tests/test_for_suite/test_for_cli/test_for_runner.py
ChristianLieven: Removed.

memote/suite/cli/runner.py
ChristianLieven: Should be fixed in the newest commit.

tests/test_for_suite/test_for_cli/test_for_runner.py
ChristianLieven: Opened #633 for this purpose! https://github.com/opencobra/memote/issues/633

PR #623, opened 2019-03-11
docs: draft guidlines for authors of custom tests

### LINKED ISSUE #559: Write up some guidelines that contribute to being able to share custom tests!
Custom tests should follow some of the principles with which we built the memote core tests.

There should be a guide in the documentation that outlines the reasoning behind some of those.

ChristianLieven:
* [X] fix #559
* [X] add entry to history.rst

I've tried drafting a set of guidelines for custom tests as per request from one of the reviewers. I'll add to this if I can think of more, but for now, let me know if you have any suggestions.

### DIFF
--- a/docs/custom_tests.rst
+++ b/docs/custom_tests.rst
@@ -215,3 +215,44 @@ module directories:
 .. code-block:: console
 
     $ memote run -a "--durations=10 path/to/dir1/ path/to/dir2/" --filename "report.html" path/to/model.xml
+
+
+Guidelines
+==========
+
+Please consider the following guidelines which reflect some of the considerations
+behind the core tests in memote. Adhering to these guidelines will allow other
+researchers to easily adopt your custom tests and ensure that they are applicable
+to a wide array of modeling practises.
+
+1. **Be namespace agnostic**. Use the ``METANETX_SHORTLIST`` and
+``COMPARTMENT_SHORTLIST`` mapping tables from memote or consider creating your
+own if your custom test needs to identify a specific metabolite in a specific
+compartment. You can generate a custom metablite shortlist by adapting
+``shortlist.tsv`` and then executing the script ``annotate_mnx_shortlists.py``
+found in ``memote/scripts``.
+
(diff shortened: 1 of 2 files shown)


docs/custom_tests.rst
```diff
@@ -215,3 +215,44 @@ module directories:
...
+
+Please consider the following guidelines which reflect some of the considerations
+behind the core tests in memote. Adhering to these guidelines will allow other
+researchers to easily adopt your custom tests and ensure that they are applicable
+to a wide array of modeling practises.
+
+1. **Be namespace agnostic**. Use the ``METANETX_SHORTLIST`` and
+``COMPARTMENT_SHORTLIST`` mapping tables from memote or consider creating your
+own if your custom test needs to identify a specific metabolite in a specific
+compartment. You can generate a custom metablite shortlist by adapting
```
kvikshaug:
```suggestion
compartment. You can generate a custom metabolite shortlist by adapting
```

docs/custom_tests.rst
```diff
@@ -215,3 +215,44 @@ module directories:
...
+
+Guidelines
+==========
+
+Please consider the following guidelines which reflect some of the considerations
+behind the core tests in memote. Adhering to these guidelines will allow other
+researchers to easily adopt your custom tests and ensure that they are applicable
+to a wide array of modeling practises.
+
+1. **Be namespace agnostic**. Use the ``METANETX_SHORTLIST`` and
```
kvikshaug: Consider including a link to where exactly this can be found (as well as `COMPARTMENT_SHORTLIST` below) so the reader doesn't have to search.

ChristianLieven: Good call! Added.

ChristianLieven:
> Point 4 and 5 seem a bit excessive - while in general good advice, is it really necessary for memote to reiterate this, or should these guidelines stick to memote- and modeling-related matters?

I think it is very important to state this cleary. If researchers will be submitting/ sharing their custom test cases, they should have an understanding of the programming standards we adhere to.

PR #634, opened 2019-03-21
test: temporarily do not expect warnings

Midnighter:
Change due to redesign in cobrapy. This will have to be reverted in
future.


ChristianLieven:
https://github.com/opencobra/memote/issues/635

PR #645, opened 2019-04-05
Git requirements

### LINKED ISSUE #524: Test for presence of git in the relevant commands

ChristianLieven:
* [x] fix #524, #523

* Include memote's requirement for git on installation instructions (README and docs).
* Check for the presence of `git` in CLI commands that require it.


memote/suite/cli/callbacks.py
```diff
@@ -97,3 +98,15 @@ def abort_if_false(ctx, param, value):
...
     if not value:
         ctx.abort()
+
+
+def git_installed():
+    """Interrupt execution of memote if `git` has not been installed."""
+    LOGGER.info("Checking `git` installation.")
+    try:
+        check_output(['git', '--version'])
+    except Exception as e:
```
Midnighter:
```suggestion
    except subprocess.CalledProcessError as error:
```

memote/suite/cli/callbacks.py
```diff
@@ -97,3 +98,15 @@ def abort_if_false(ctx, param, value):
...
+        LOGGER.critical(
+            "The execution of memote was interrupted since no installation of "
+            "`git` could be detected. {}".format(str(e)))
```
Midnighter: I think in this case we don't need to attach the exact error message. If you do want to make it possible to show the message I would add below a debug message (see code below). Instead I would add a URL for how to install git, for example, https://git-scm.com/book/en/v2/Getting-Started-Installing-Git

```python
LOGGER.debug("Underlying error:", exc_info=error)
```

memote/suite/cli/callbacks.py
ChristianLieven: Done!

memote/suite/cli/callbacks.py
ChristianLieven: Done!

PR #662, opened 2019-04-28
Have memote.utils.is_modified() return False if file was deleted.

jonovik:
`memote history` will fail on a commit that deleted the model file.

```
Running the test suite for commit '8daf...'.
Traceback (most recent call last):
...
  File "C:\NMBU\Miniconda3\envs\cbm\lib\site-packages\click\core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "c:\git\digisal\memote\memote\suite\cli\runner.py", line 393, in history
    blob = cmt.tree[model]
  File "C:\NMBU\Miniconda3\envs\cbm\lib\site-packages\git\objects\tree.py", line 298, in __getitem__
    return self.join(item)
  File "C:\NMBU\Miniconda3\envs\cbm\lib\site-packages\git\objects\tree.py", line 225, in join
    item = tree[token]
  File "C:\NMBU\Miniconda3\envs\cbm\lib\site-packages\git\objects\tree.py", line 298, in __getitem__
    return self.join(item)
  File "C:\NMBU\Miniconda3\envs\cbm\lib\site-packages\git\objects\tree.py", line 244, in join
    raise KeyError(msg % file)
KeyError: "Blob or Tree named 'export' not found"
```

(where `export` referred to my model file `export/model.xml`)

This is because [memote.utils.is_modified](https://github.com/opencobra/memote/blob/276630fcd4449fb7b914186edfd38c239e7052df/memote/utils.py#L269) just tests for `path in commit.stats.files`, whereas a deletion shows up as `{'insertions': 0, 'deletions': N, 'lines': N}` where `N` is the number of lines in the file.

[This example](https://github.com/jonovik/memote/blob/37fdb6a4b58952d57a9c78c0a31cba1a4757c321/tests/test_for_utils.py#L125) shows a file being created, modified, untouched, deleted (in reverse chronological order):

```python
>>> [commit.stats.files for commit in repo.iter_commits()]
[{'test_file.txt': {'insertions': 0, 'deletions': 1, 'lines': 1}},
 {},
 {'test_file.txt': {'insertions': 1, 'deletions': 1, 'lines': 2}},
 {'test_file.txt': {'insertions': 1, 'deletions': 0, 'lines': 1}}]
```

Since we cannot run tests on a nonexistent model file, it is more robust to skip those commits.
I have modified [memote.utils.is_modified](https://github.com/jonovik/memote/blob/37fdb6a4b58952d57a9c78c0a31cba1a4757c321/memote/utils.py#L269) and added a [test](https://github.com/jonovik/memote/blob/37fdb6a4b58952d57a9c78c0a31cba1a4757c321/tests/test_for_utils.py#L147) which fails in 276630f but passes in my pull request.

### DIFF
--- a/tests/test_for_utils.py
+++ b/tests/test_for_utils.py
@@ -135,8 +135,10 @@ def mock_repo(tmpdir_factory):
         repo.index.add([relname])
         repo.index.commit(message)
 
-    repo.index.add([relname])
-    repo.index.commit("Empty commit.")
+    with open(os.path.join(path, "unrelated_file.txt"), "w") as f:
+        f.write("Unrelated file.")
+    repo.index.add(["unrelated_file.txt"])
+    repo.index.commit("A commit that does not touch {}".format(relname))
 
     repo.index.remove([relname], working_tree=True)
     repo.index.commit("Delete file.")
--- a/tests/test_for_utils.py
+++ b/tests/test_for_utils.py
@@ -124,7 +124,7 @@ def test_show_versions(capsys):
 @pytest.fixture(scope="function")
 def mock_repo(tmpdir_factory):
     """Mock repo with history: add, modify, do nothing, delete file."""
-    path = tmpdir_factory.mktemp("memote-mock-repo")
+    path = str(tmpdir_factory.mktemp("memote-mock-repo"))
     repo = git.Repo.init(path)
     relname = "test_file.txt"
     absname = os.path.join(path, relname)
--- a/tests/test_for_utils.py
+++ b/tests/test_for_utils.py
@@ -15,7 +15,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
+
 import pytest
+import git
 
 import memote.utils as utils
 
@@ -116,3 +119,40 @@ def test_show_versions(capsys):
     assert lines[7].startswith("Package Versions")
     assert lines[8].startswith("================")
     assert any(l.startswith("memote") for l in lines[9:])
+
+
+@pytest.fixture(scope="function")
+def mock_repo(tmpdir_factory):
+    """Mock repo with history: add, modify, do nothing, delete file."""
+    path = tmpdir_factory.mktemp("memote-mock-repo")
+    repo = git.Repo.init(path)
+    relname = "test_file.txt"
+    absname = os.path.join(path, relname)
+
+    for message in "Add file.", "Modify file.":
+        with open(absname, "w") as f:
+            f.write(message)
+        repo.index.add([relname])
+        repo.index.commit(message)
+
+    repo.index.add([relname])
+    repo.index.commit("Empty commit.")
+
+    repo.index.remove([relname], working_tree=True)
+    repo.index.commit("Delete file.")
+
+    return relname, repo
+
+
+def test_is_modified_deleted(mock_repo):
+    """Don't report file deletion as modification."""
+    relname, repo = mock_repo
+    # File history (newest first): deleted, unchanged, modified, created
+    # Don't report file deletion as "modification"
+    # for purposes of running memote tests on the file.
+    want = False, False, True, True
+    # Convert to tuple so we can compare want == got,
+    # which pytest will introspect helpfully if the assertion fails.
+    got = tuple(utils.is_modified(relname, commit)
+                for commit in repo.iter_commits())
+    assert want == got
--- a/tests/test_for_utils.py
+++ b/tests/test_for_utils.py
@@ -126,6 +126,9 @@ def mock_repo(tmpdir_factory):
     """Mock repo with history: add, modify, do nothing, delete file."""
     path = str(tmpdir_factory.mktemp("memote-mock-repo"))
     repo = git.Repo.init(path)
+    with repo.config_writer() as writer:
+        writer.set_value("user", "name", "memote-bot").release()
+        writer.set_value("user", "email", "bot@memote.io").release()
     relname = "test_file.txt"
     absname = os.path.join(path, relname)
 
@@ -133,15 +136,18 @@ def mock_repo(tmpdir_factory):
         with open(absname, "w") as f:
             f.write(message)
         repo.index.add([relname])
-        repo.index.commit(message)
+        repo.git.commit("--message", message)
 
     with open(os.path.join(path, "unrelated_file.txt"), "w") as f:
         f.write("Unrelated file.")
     repo.index.add(["unrelated_file.txt"])
-    repo.index.commit("A commit that does not touch {}".format(relname))
+    repo.git.commit(
+        "--message",
+        "A commit that does not touch {}".format(relname)
+    )
 
     repo.index.remove([relname], working_tree=True)
-    repo.index.commit("Delete file.")
+    repo.git.commit("--message", "Delete file.")
 
     return relname, repo
 
--- a/scripts/update_mock_repo.py
+++ b/scripts/update_mock_repo.py
@@ -27,16 +27,17 @@ from __future__ import absolute_import
 
 import logging
 import os
+import tarfile
 from os.path import join
 from shutil import rmtree
-from subprocess import check_output, call
 from tempfile import mkdtemp
-import tarfile
 
-logging.basicConfig(level=logging.INFO)
+import git
+
 LOGGER = logging.getLogger()
 
-def update_mock_repo():
+
+def update_mock_repo(temp_dir):
     """
     Clone and gzip the memote-mock-repo used for CLI and integration tests.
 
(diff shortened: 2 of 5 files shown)


memote/utils.py
```diff
@@ -280,10 +280,19 @@ def is_modified(path, commit):
...
     -------
     bool
-        Whether or not the given path is among the modified files.
+        Whether or not the given path is among the modified files
+        and was not deleted in this commit.
 
     """
-    return path in commit.stats.files
+    try:
+        d = commit.stats.files["path"]
```
Midnighter: I think this should be the `path` variable, not a string. Also, I prefer a more descriptive variable name. Maybe `changes`?

```suggestion
        changes = commit.stats.files[path]
```

memote/utils.py
```diff
@@ -280,10 +280,19 @@ def is_modified(path, commit):
...
+        if (d["insertions"] == 0) and (d["deletions"] == d["lines"]):
```
Midnighter:
```suggestion
        if (changes["insertions"] == 0) and (changes["deletions"] == changes["lines"]):
```

tests/test_for_utils.py
```diff
@@ -116,3 +119,40 @@ def test_show_versions(capsys):
...
+    relname = "test_file.txt"
+    absname = os.path.join(path, relname)
+
+    for message in "Add file.", "Modify file.":
+        with open(absname, "w") as f:
+            f.write(message)
+        repo.index.add([relname])
+        repo.index.commit(message)
+
+    repo.index.add([relname])
```
Midnighter: I'm surprised this creates a commit at all. Usually adding an unmodified file has no effect and git won't let you make a commit without any changes. Maybe create and add another unrelated file here?

jonovik:
> I think the tests failing are due to the typo with the `path` variable. So hopefully it's an easy fix. Thanks a lot for spotting this and providing a solution 😃

That is curious. The tests pass on my system; see below.

Just to be clear: `test_for_utils.py::test_is_modified_deleted` was designed to fail in the current memote, demonstrating the problem. And with my modifications to `memote.utils.is_modified`, it passes.

I was waiting for continuous integration to verify that it also works under linux, but am not intimately familiar with Travis. Are the "Some checks were not successful" referring to my commit, or to your proposed changes?

```
(cbm) C:\git\digisal\memote\tests>pytest -v test_for_utils.py
======================================================================================================== test session starts ========================================================================================================
platform win32 -- Python 3.6.8, pytest-4.4.1, py-1.8.0, pluggy-0.9.0 -- c:\nmbu\miniconda3\envs\cbm\python.exe
cachedir: .pytest_cache
metadata: {'Python': '3.6.8', 'Platform': 'Windows-10-10.0.17763-SP0', 'Packages': {'pytest': '4.4.1', 'py': '1.8.0', 'pluggy': '0.9.0'}, 'Plugins': {'metadata': '1.8.0', 'html': '1.20.1.dev1+g27b6e77'}, 'JAVA_HOME': 'C:\\Program Files\\Java\\jdk1.8.0_112\\jre'}
rootdir: C:\git\digisal\memote, inifile: setup.cfg
plugins: metadata-1.8.0, html-1.20.1.dev1+g27b6e77
collected 11 items

test_for_utils.py::test_register_with[one-one] PASSED                                                                                                                                                                          [  9%]
test_for_utils.py::test_register_with[two-two] PASSED                                                                                                                                                                          [ 18%]
test_for_utils.py::test_register_with[three-three] PASSED                                                                                                                                                                      [ 27%]
test_for_utils.py::test_register_with[four-four] PASSED                                                                                                                                                                        [ 36%]
test_for_utils.py::test_annotate[notes0-one-One line.] PASSED                                                                                                                                                                  [ 45%]
test_for_utils.py::test_annotate[notes1-two-Two lines.\n    Why?\n    ] PASSED                                                                                                                                                 [ 54%]
test_for_utils.py::test_annotate[notes2-three-\n    Three lines.\n    Why?\n    ] PASSED                                                                                                                                       [ 63%]
test_for_utils.py::test_annotate[notes3-four-\n    Fourth summary.\n\n    A proper description.\n\n    Returns\n    -------\n    None\n\n    ] PASSED                                                                          [ 72%]
test_for_utils.py::test_annotate[notes4-one-] FAILED                                                                                                                                                                           [ 81%]
test_for_utils.py::test_show_versions PASSED                                                                                                                                                                                   [ 90%]
test_for_utils.py::test_is_modified_deleted PASSED
```

jonovik:
@Midnighter , I'm sorry. I had forgotten to *commit* my latest changes. Hang on and we'll see if it works out...

jonovik:
> @Midnighter , I'm sorry. I had forgotten to _commit_ my latest changes. Hang on and we'll see if it works out...

Please disregard that comment. The uncommitted changes were a reversal of memote.utils while keeping my tests.test_for_utils, which I had used to verify that my test was indeed failing before.

After reinstalling memote from that version (with my test but the buggy utils.py) with 
pip install --upgrade c:/git/digisal/memote
I am getting the failure that I expected; see below.

However, I note that the failure is different than what Travis CI is showing:  [`E       assert (False, False, True, True) == (False, False, False, False)`](https://travis-ci.org/opencobra/memote/jobs/525533526#L627)  
versus my `E       assert (False, False, True, True) == (True, False, True, True)`.

So maybe it is gitpython behaving differently under Windows (this is in the usual CMD shell, not git bash). I will zip up my mock repo and email it to you. I'll also push your suggestion of including an unrelated file in the mock commit as you proposed, in case that makes behaviour more reproducible.

```
(cbm) C:\git\digisal\memote>pytest -v tests/test_for_utils.py
======================================================================================================== test session starts ========================================================================================================
platform win32 -- Python 3.6.8, pytest-4.4.1, py-1.8.0, pluggy-0.9.0 -- c:\nmbu\miniconda3\envs\cbm\python.exe
cachedir: .pytest_cache
metadata: {'Python': '3.6.8', 'Platform': 'Windows-10-10.0.17763-SP0', 'Packages': {'pytest': '4.4.1', 'py': '1.8.0', 'pluggy': '0.9.0'}, 'Plugins': {'metadata': '1.8.0', 'html': '1.20.1.dev1+g27b6e77'}, 'JAVA_HOME': 'C:\\Program Files\\Java\\jdk1.8.0_112\\jre'}
rootdir: C:\git\digisal\memote, inifile: setup.cfg
plugins: metadata-1.8.0, html-1.20.1.dev1+g27b6e77
collected 11 items
...
tests/test_for_utils.py::test_show_versions PASSED                                                                                                                                                                             [ 90%]
tests/test_for_utils.py::test_is_modified_deleted FAILED                                                                                                                                                                       [100%]

============================================================================================================= FAILURES ==============================================================================================================
...
c:\nmbu\miniconda3\envs\cbm\lib\site-packages\memote\utils.py:119: ValueError
_____________________________________________________________________________________________________ test_is_modified_deleted ______________________________________________________________________________________________________

mock_repo = ('test_file.txt', <git.Repo "C:\Users\jonvi\AppData\Local\Temp\pytest-of-jonvi\pytest-3\memote-mock-repo0\.git">)

    def test_is_modified_deleted(mock_repo):
        """Don't report file deletion as modification."""
        relname, repo = mock_repo
        # File history (newest first): deleted, unchanged, modified, created
        # Don't report file deletion as "modification"
        # for purposes of running memote tests on the file.
        want = False, False, True, True
        # Convert to tuple so we can compare want == got,
        # which pytest will introspect helpfully if the assertion fails.
        got = tuple(utils.is_modified(relname, commit)
                    for commit in repo.iter_commits())
>       assert want == got
E       assert (False, False, True, True) == (True, False, True, True)
E         At index 0 diff: False != True
E         Full diff:
E         - (False, False, True, True)
E         + (True, False, True, True)

tests\test_for_utils.py:158: AssertionError
================================================================================================ 2 failed, 9 passed in 3.80 seconds =================================================================================================
```

jonovik:
> So maybe it is gitpython behaving differently under Windows (this is in the usual CMD shell, not git bash). I will zip up my mock repo and email it to you. I'll also push your suggestion of including an unrelated file in the mock commit as you proposed, in case that makes behaviour more reproducible.

@Midnighter Adding an unrelated file does indeed give the same error that you are seeing. I'll look into why that would break the logic of my `is_modified()`. I'll ping you when I have a solution.

jonovik:
@Midnighter You were right about the `path` typo. What baffles me is how this could still be passing on my computer. Hope everything is okay now. (Tests pass on my system, for what that's worth 8-)

memote/utils.py
jonovik: Sorry, I forgot to consider/comment on the proposed name change. Since the dictionary `d` is used three times in the same expression and self-explanatorily defined in the preceding line, I think this is clearer than having to break the code over another line.

tests/test_for_utils.py
jonovik: This is a documented feature of git, but as the [docs](https://git-scm.com/docs/git-commit) say, mostly used for scripting conversion from other SCMs. I use it when reconstructing a history containing only xml and yml exports of the multi-file proprietary model format we've been struggling with.

git commit --allow-empty -m "Initial commit intentionally left blank."

jonovik:
I see that the build is [failing on Python 3.5](https://travis-ci.org/opencobra/memote/jobs/525806677#L616) and presumably earlier versions due to something something LocalPath. I'll have another look at the git repo mockup I was imitating.

Midnighter:
Looking at the Travis output, I think you just need a small change in the fixture.

tests/test_for_utils.py
```diff
@@ -116,3 +119,42 @@ def test_show_versions(capsys):
     assert lines[7].startswith("Package Versions")
     assert lines[8].startswith("================")
     assert any(l.startswith("memote") for l in lines[9:])
+
+
+@pytest.fixture(scope="function")
+def mock_repo(tmpdir_factory):
+    """Mock repo with history: add, modify, do nothing, delete file."""
+    path = tmpdir_factory.mktemp("memote-mock-repo")
```
Midnighter: I believe this change should fix the failing tests.

```suggestion
    path = str(tmpdir_factory.mktemp("memote-mock-repo"))
```

jonovik:
> Looking at the Travis output, I think you just need a small change in the fixture.

Yes, I think I know what to do (convert LocalPath to str). Just waiting for conda to make me a 3.5 test environment.

Midnighter:
Ahh, sorry @jonovik. @ChristianLieven reworked a lot of the git interaction logic recently because gitpython doesn't fully work on Windows. If you want we can take it from here. Thank you for the patch and test so far!

jonovik:
> Ahh, sorry @jonovik. @ChristianLieven reworked a lot of the git interaction logic recently because gitpython doesn't fully work on Windows. If you want we can take it from here. Thank you for the patch and test so far!

@Midnighter Please do take it from here.

@ChristianLieven As for git on Windows, I've found that `repo.git.commit` works where `repo.index.commit` fails in getting [hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) to work on Windows. I am eagerly awaiting your reworking. Here are a couple of my own notes, in case they may be useful (commit message from a private branch):

> **Use `repo.git.commit` instead of `repo.index.commit` for hooks to work on Windows.**
> 
> Given `repo = git.Repo(...)`, Gitpython's `repo.index.commit(message)` uses `os.subprocess` to start `.git/hook/pre-commit`.
> However, Windows does not recognize this text file as a Python script, and so this fails with
> `OSError: [WinError 193] %1 is not a valid Win32 application`
> 
> `Using repo.git.commit("-m", message)` instead invokes the `git` executable instead, which (I think) parses the shebang of `.git/hook/pre-commit` and does the right thing.

Midnighter:
Made some more modifications in order to standardize git use across the project. Thank you very much for this addition @jonovik and my sincere apologies for dragging this out so long. It fails on Travis for Python 2.7 but works for me locally. Since we are anyway dropping support for 2, I will merge it here.

PR #668, opened 2019-06-21
chore: upgrade structure and configuration

### LINKED ISSUE #665: Support latest cobrapy
#### Problem description
I have some models which I can only load with the latest cobrapy 0.15.3 (mainly SBML level 3 version 2). It would be great if memote could run with the latest cobrapy version (or at least having a branch which can).
I would like to run memote as part of a continuous model development pipeline, but would require latest library versions of cobrapy, pandas, ...

Necessary to implement the changes to support the latest cobrapy (and pandas).

Midnighter:
* [x] fix #665 
* [x] tests added/passed


src/memote/experimental/essentiality.py
```diff
@@ -87,7 +87,6 @@ def evaluate(self, model):
             essen = single_gene_deletion(
                 model, gene_list=self.data["gene"], processes=1)
         essen["gene"] = [list(g)[0] for g in essen.index]
-        essen.index = essen["gene"]
```
ChristianLieven: Is this change to satisfy the updated pandas?

Midnighter: Yes, it now complains when the index and a column have the same name. Another solution would be to just give it a different name.

PR #670, opened 2019-08-05
chore: register custom pytest markers to snuff error messages

ChristianLieven:

This removes the warnings that appear with the most recent version of Pytest under Python 3.6.0 by registering the custom markers we use for parametrization (`biomass`, `essentiality` and `growth`). 

Running `memote report snapshot` on Google Colaboratory (Python 3.6.8) errors with something along these lines:
```'biomass' not found in `markers` configuration option```

This is [recommended by the authors of pytest](https://docs.pytest.org/en/latest/example/markers.html?highlight=register%20markers#registering-markers)

### DIFF
--- a/src/memote/suite/collect.py
+++ b/src/memote/suite/collect.py
@@ -170,7 +170,7 @@ class ResultCollectionPlugin(object):
         return self._sbml_ver
 
     def pytest_configure(self, config):
-        """Register custom markers at runtime """
+        """Register custom markers at runtime."""
         config.addinivalue_line("markers",
                                 "biomass: ")
         config.addinivalue_line("markers",
--- a/src/memote/suite/collect.py
+++ b/src/memote/suite/collect.py
@@ -168,3 +168,12 @@ class ResultCollectionPlugin(object):
     def sbml_version(self):
         """Provide SBML level, version, and FBC use."""
         return self._sbml_ver
+
+    def pytest_configure(self, config):
+        """Register custom markers at runtime """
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/memote/suite/collect.py [hunk 1]
```diff
@@ -168,3 +168,12 @@ def model(self):
     def sbml_version(self):
         """Provide SBML level, version, and FBC use."""
         return self._sbml_ver
+
+    def pytest_configure(self, config):
+        """Register custom markers at runtime."""
+        config.addinivalue_line("markers",
+                                "biomass: ")
```
Midnighter: Usually markers are only the word itself, are you sure this is how it should be?

```suggestion
                                "biomass")
```

ChristianLieven: It's possible that it is not required, however, all examples I found were in the format:
```
  def pytest_configure(self, config):
        """Register custom markers at runtime."""
        config.addinivalue_line("markers",
                                "marker: description of marker")
```

I'm happy to try it without the colon.

Midnighter:
We probably have to drop Python 2 support since the scientific Python dependencies have already done so.

ChristianLieven:
>drop py2 support

The sooner the better! I'm happy to make a quick and dirty switchover ASAP (import warning and removal of py2 CI). We should probably discuss steps on how to do a full transition (similar to solution 3 suggested here: https://github.com/opencobra/cobrapy/issues/853#issuecomment-515514559)

PR #671, opened 2019-08-30
Deprecate py2

ChristianLieven:


.travis.yml
```diff
@@ -3,9 +3,9 @@ node_js: "node"
 os:
   - linux
 python:
+  - "3.7"
```
Midnighter: I think 3.7 is still not standard on Travis.

```suggestion
  - "3.7-dev"
```

README.rst
```diff
@@ -61,6 +61,14 @@ at via GitHub pages for your repository.
 Installation
 ============
 
+**Please Note**: With its
+`retirement on the horizon <https://pythonclock.org>`_ we decided to stop
+testing against Python 2.7,
+and `like many others <https://python3statement.org>`_, want to focus entirely
+on Python 3.
+Hence, we cannot guarantee that the `memote` will still function as expected
```
Midnighter:
```suggestion
Hence, we cannot guarantee that `memote` will still function as expected
```

setup.cfg
```diff
@@ -70,7 +67,7 @@ install_requires =
 	depinfo
 	requests
 	equilibrator_api <0.2;python_version>='3.5'
-python_requires = >=2.7,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*
+python_requires = >=3.5,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*
```
Midnighter:
```suggestion
python_requires = >=3.5
```

Midnighter: This will prevent installation on 2.7, though.

ChristianLieven: Yeah I was wondering if we needed that :D

src/memote/__init__.py
```diff
@@ -30,6 +30,18 @@
...
 from memote.jinja2_extension import *
 
+import sys
+import logging
+
 __author__ = "Moritz E. Beber"
 __email__ = "morbeb@biosustain.dtu.dk"
 __version__ = "0.9.12"
+
+LOGGER = logging.getLogger()
```
Midnighter: It's still logging within memote so I prefer
```suggestion
LOGGER = logging.getLogger(__name__)
```

src/memote/__init__.py
```diff
@@ -30,6 +30,18 @@
...
+
+if sys.version_info < (3,):
+    LOGGER.warning('You are using `Memote` with Python 2. '
```
Midnighter: Rather than logging a warning I would raise a `FutureWarning`.

.travis.yml
ChristianLieven: The build works on 3.7, should I then just leave it as is? Or is the `3.7-dev` special in a way beyond that?

src/memote/__init__.py
ChristianLieven: It doesn't look like this check runs when memote is installed with `pip > 9.0` AND `python_requires = >=3.5`. In that case, pip will actually already print the following error when installing memote:

```
ERROR: Package u'memote' requires a different Python: 2.7.9 not in '>=3.5,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*'
```

ChristianLieven: But I suppose it would if users manage to install memote with an older version of pip.

src/memote/__init__.py
ChristianLieven: Resolving since I removed the logging altogether in favor of a warning as per your advice below.

setup.cfg
Midnighter: I think for now we should at least still allow installation on Python 2.7. So maybe just leave this line unchanged.

.travis.yml
Midnighter: `-dev` installs the latest release candidate so it is a bit different. However, if 3.7 now exists Please ignore this suggestion :smiley:

PR #677, opened 2019-10-19
fix: change biomass matching logic

### LINKED ISSUE #675: Memote warns about the existence of two biomass reactions, while there is only one.
Opening on behalf of @anushcph

It might be a good idea to include other criteria like at least a number of biomass precursors as part of the reaction or the “biomass” metabolite should only be product in the reaction

Midnighter:
* [x] fix #675 
* [x] tests added/passed

### DIFF
--- a/src/memote/support/helpers.py
+++ b/src/memote/support/helpers.py
@@ -22,8 +22,7 @@ from __future__ import absolute_import
 import logging
 import re
 from collections import defaultdict
-from operator import itemgetter
-
+from operator import itemgetter, attrgetter
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


src/memote/support/helpers.py
```diff
@@ -278,21 +277,59 @@ def find_converting_reactions(model, pair):
...
-    5. Return the union of reactions that match the buzzwords and of the
-    reactions that metabolites are involved in that match the buzzword.
+    Identifiy possible biomass reactions using multiple steps:
+    1. Look for candidate reactions that include the SBO term ``SBO:0000629``
+    for biomass production,
+    2. the 'buzzwords' biomass, growth, and bof in reaction names
+    and identifiers,
+    3. reactions that involve a metabolite with the SBO term ``SBO:0000649``
+    for biomass,
+    3. or reactions that involve a metabolite whose name or identifier
```
ChristianLieven: Small change: `3.` should be `4.` right.

PR #678, opened 2019-10-31
fix: change logic of test production and consumption metabolites to speed up the process

carrascomj:
In order to speed up the consistency tests of checking metabolite production and consumption with open bounds, the logic of the test has been changed. Instead of adding a boundary to the model each time, the test now works by sequentially changing the linear coefficients of the metabolites, which should be faster.

* Add a variable **just once** (bounds in 0,1000);
* Set the variable as the objective of the model;
* for each metabolite:  
        ▪ get the linear coefficients of the constraint of each metabolite;  
        ▪ add a coefficient to the variable in the constraint (-1 for production and 1 for consumption)
        ▪ solve;  
        ▪ restore constraint;  
        ▪ If not solution, add the metabolite to the returned output.

**Multiprocessing** was also added to decrease the running time.

### DIFF
--- a/src/memote/support/consistency.py
+++ b/src/memote/support/consistency.py
@@ -623,7 +623,7 @@ def find_metabolites_not_consumed_with_open_bounds(model, processes=None):
         Those metabolites that could not be consumed.
 
     """
-    find_metabolites_not_produced_with_open_bounds(
+    return find_metabolites_not_produced_with_open_bounds(
         model, processes=processes, prod=False
     )
 
--- a/src/memote/support/consistency.py
+++ b/src/memote/support/consistency.py
@@ -492,7 +492,7 @@ def find_disconnected(model):
     return [met for met in model.metabolites if len(met.reactions) == 0]
 
 
-def _init_worker(model, irr):
+def _init_worker(model, irr, val):
     """Initialize a global model object for multiprocessing.
 
     Parameters
@@ -502,12 +502,16 @@ def _init_worker(model, irr):
     irr: optlang.Variable || cobra.Reaction
         the reaction to be added to the linear coefficients. It must be in the
         variables of the model.
+    val: int
+        value of the coefficient: -1 for production and 1 for consumption
     """
     global _model
     global _irr
+    global _val
     _model = model
     _model.objective = irr
     _irr = irr
+    _val = val
 
 
 def _solve_metabolite_production(metabolite):
@@ -529,14 +533,15 @@ def _solve_metabolite_production(metabolite):
     """
     global _model
     global _irr
+    global _val
     constraint = _model.metabolites.get_by_id(metabolite.id).constraint
-    constraint.set_linear_coefficients({_irr: -1})
+    constraint.set_linear_coefficients({_irr: _val})
     solution = _model.slim_optimize()
     constraint.set_linear_coefficients({_irr: 0})
     return solution, metabolite
 
 
-def find_metabolites_not_produced_with_open_bounds(model, processes=None):
+def find_metabolites_not_produced_with_open_bounds(model, processes=None, prod=True):
     """
     Return metabolites that cannot be produced with open exchange reactions.
 
@@ -552,6 +557,8 @@ def find_metabolites_not_produced_with_open_bounds(model, processes=None):
         The metabolic model under investigation.
     processes: int
         Number of processes to be used (Default to `cobra.Configuration()`).
+    prod: bool
+        If False, it checks for consumption instead of production (Default True)
 
     Returns
     -------
@@ -564,6 +571,10 @@ def find_metabolites_not_produced_with_open_bounds(model, processes=None):
         processes = Configuration().processes
     n_mets = len(model.metabolites)
     processes = min(processes, n_mets)
+    # manage the value of the linear coefficient to be added to each metabolite
+    val = -1 # production
+    if not prod:
+        val = 1 # consumption
 
     helpers.open_exchanges(model)
     irr = model.problem.Variable("irr", lb=0, ub=1000)
@@ -573,7 +584,7 @@ def find_metabolites_not_produced_with_open_bounds(model, processes=None):
         pool = multiprocessing.Pool(
             processes,
             initializer=_init_worker,
-            initargs=(model, irr),
+            initargs=(model, irr, val),
         )
         # use map as filter
         mets_not_produced = [met for solution, met in pool.imap_unordered(
@@ -590,7 +601,7 @@ def find_metabolites_not_produced_with_open_bounds(model, processes=None):
     return mets_not_produced
 
 
-def find_metabolites_not_consumed_with_open_bounds(model):
+def find_metabolites_not_consumed_with_open_bounds(model, processes=None):
     """
     Return metabolites that cannot be consumed with open boundary reactions.
 
@@ -601,6 +612,8 @@ def find_metabolites_not_consumed_with_open_bounds(model):
     ----------
     model : cobra.Model
         The metabolic model under investigation.
+    processes: int
+        Number of processes to be used (Default to `cobra.Configuration()`).
 
     Returns
     -------
@@ -608,16 +621,7 @@ def find_metabolites_not_consumed_with_open_bounds(model):
         Those metabolites that could not be consumed.
 
     """
-    mets_not_consumed = list()
-    helpers.open_exchanges(model)
-    for met in model.metabolites:
-        with model:
-            exch = model.add_boundary(
-                met, type="irrex", reaction_id="IRREX", lb=-1000, ub=0)
-            solution = helpers.run_fba(model, exch.id, direction="min")
-            if np.isnan(solution) or abs(solution) < TOLERANCE_THRESHOLD:
-                mets_not_consumed.append(met)
-    return mets_not_consumed
+    find_metabolites_not_produced_with_open_bounds(model, processes=processes, prod=False)
 
 
 def find_reactions_with_unbounded_flux_default_condition(model):
--- a/src/memote/support/consistency.py
+++ b/src/memote/support/consistency.py
@@ -535,9 +535,6 @@ def _solve_metabolite_production(metabolite):
         metabolite passed as argument (to use map as a filter)
 
     """
-    global _model
-    global _irr
-    global _val
     constraint = _model.metabolites.get_by_id(metabolite.id).constraint
     constraint.set_linear_coefficients({_irr: _val})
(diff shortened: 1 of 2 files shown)


ChristianLieven:
I like your changes especially the use of mutliprocessing and manipulating the underlying model directly.

I'd be interested to see how much faster this is. Would you mind providing a before/after comparison?

And, of course, I'd feel more comfortable merging it if the test wouldn't fail.

carrascomj:
The tests should pass now. I have the comparison in a notebook, but apparently GitHub doesn't support the format. I'll try to paste the code and the results.

carrascomj:
The tests were performed on a small model in BiGG.
```
![ ! -f "iAB_RBC_283.xml" ] && curl -L -O 'http://bigg.ucsd.edu/static/models/iAB_RBC_283.xml'
VERBOSITY = False
PROCESSES = 4
```

# 1. Test change in logic
Old functions: 

```python
from time import time

import cobra
import numpy as np
from copy import deepcopy
from cobra.exceptions import Infeasible
from tqdm import tqdm
import sys

def run_fba(model, rxn_id, direction="max", single_value=True):
    model.objective = model.reactions.get_by_id(rxn_id)
    model.objective_direction = direction
    if single_value:
        return model.slim_optimize()
    else:
        try:
            solution = model.optimize()
        except Infeasible:
            return solution
            return np.nan


def open_exchanges(model):
    for rxn in model.exchanges:
        rxn.bounds = (-1000, 1000)


def test_old(model):
    """
    old_test in consistency.py
    """
    mets_not_produced = list()
    open_exchanges(model)
    pbar = tqdm(total=len(model.metabolites))
    for met in model.metabolites:
        with model:
            exch = model.add_boundary(
                met, type="irrex", reaction_id="IRREX", lb=0, ub=1000
            )
            solution = run_fba(model, exch.id)
            if np.isnan(solution) or solution < model.tolerance:
                mets_not_produced.append(met)
        pbar.update(1)
    return mets_not_produced
```
New logic functions:

```python
def solve_boundary(metabolite, rxn, val=-1):
    """
    Solves the model when some reaction `rxn` has been added to the `metabolite`'s contraints.
    """
    metabolite.constraint.set_linear_coefficients({rxn: val})
    solution = metabolite.model.slim_optimize()
    # TODO: it seems like with context doesn't catch these changes, need to check
    # restore constraint
    metabolite.constraint.set_linear_coefficients({rxn: 0})
    return solution


def test_new(model):
    """
    New test
    """
    mets_not_produced = list()
    open_exchanges(model)
    irr = model.problem.Variable("irr", lb=0, ub=1000)
    with model:
        model.add_cons_vars(irr)
        # helper.run_fba() only accepts reactions in the model
        model.objective = irr
        pbar = tqdm(total=len(model.metabolites))
        for met in model.metabolites:
            solution = solve_boundary(met, irr)
            if np.isnan(solution) or solution < model.tolerance:
                mets_not_produced.append(met)
            pbar.update(1)
    return mets_not_produced
```
Comparison:

```python
model = cobra.io.read_sbml_model("iAB_RBC_283.xml")
model.solver = "glpk"
if VERBOSITY:
    model.solver.interface.Configuration.verbosity = model.solver.interface.Configuration(
        verbosity=3
    )

print(f"Number of metabolites in the model: {len(model.metabolites)}")
start = time()
old = set(test_old(model))
print(
    f"Identified metabolites by the old version: {len(old)} in {time() - start} s"
)
# old is 99 for this model in 12.536982536315918 s
start = time()
new = set(test_new(model))
print(
    f"Identified metabolites by the new version: {len(new)} in {time() - start} s"
)
print(old ^ new)

assert old == new
```
### Results
```
Number of metabolites in the model: 342
Identified metabolites by the old version: 22 in 2.8081214427948 s
Identified metabolites by the new version: 22 in 2.1633551120758057 s
set()
```
It's not a huge difference; although profiling revealed the expected behavior.

# 2. Multiprocessing

Functions for multiprocessing:

```python
import multiprocessing

def _init_worker(model, irr, val):
    """Initialize a global model object for multiprocessing.

    Parameters
    ----------
    model : cobra.Model
        The metabolic model under investigation.
    irr: optlang.Variable || cobra.Reaction
        the reaction to be added to the linear coefficients. It must be in the
        variables of the model.
    val: int
        value of the linear coefficient (1 for consumption, -1 for production)
    """
    global _model
    global _irr
    global _val
    _model = model
    _model.objective = irr
    _irr = irr
    _val = val


def _solve_metabolite_production(metabolite):
    """
    Solves the model when some reaction has been added to a `metabolite`'s
    contraints. The reaction and the model are passed as globals.

    Parameters
    ----------
    metabolite: cobra.Metabolite
        the reaction will be added to this metabolite as a linear coefficient

    Returns
    -------
    solution: float
        the value of the solution of the LP problem, *NaN* if infeasible.
    metabolite: cobra.Metabolite
        metabolite passed as argument (to use map as a filter)
    """
    global _model
    global _irr
    global _val
    constraint = _model.metabolites.get_by_id(metabolite.id).constraint
    constraint.set_linear_coefficients({_irr: _val})
    solution = _model.slim_optimize()
    constraint.set_linear_coefficients({_irr: 0})
    return solution, metabolite


def find_metabolites_not_produced_with_open_bounds(model, processes=None, prod = True):
    """
    Return metabolites that cannot be produced with open exchange reactions.

    A demand reaction is set as the objective. Then, it is sequentally added as
    a coefficient for every metabolite and the solution is inspected.

    A perfect model should be able to produce each and every metabolite when
    all medium components are available.

    Parameters
    ----------
    model : cobra.Model
        The metabolic model under investigation.
    processes: int
        Number of processes to be used (Default to `cobra.Configuration()`).
    prod: bool
        If False, it checks for consumption instead of production. Default True

    Returns
    -------
    list
        Those metabolites that could not be produced.

    """
    if processes is None:
        # For now, borrow the number of processes from cobra's configuration
        processes = Configuration().processes
    n_mets = len(model.metabolites)
    processes = min(processes, n_mets)
    # manage the value of the linear coefficient to be added to each metabolite
    val = -1 # production
    if not prod:
        val = 1 # consumption
    open_exchanges(model)
    irr = model.problem.Variable("irr", lb=0, ub=1000)

    if processes > 1:
        chunk_s = n_mets // processes
        pool = multiprocessing.Pool(
            processes,
            initializer=_init_worker,
            initargs=(model, irr, val),
        )
        # use map as filter
        mets_not_produced = [met for solution, met in pool.imap_unordered(
            _solve_metabolite_production, model.metabolites, chunksize=chunk_s
        ) if np.isnan(solution) or solution < model.tolerance]
        pool.close()
        pool.join()
    else:
        _init_worker(model, irr)
        # use map as filter
        mets_not_produced = [met for solution, met in map(
            _solve_metabolite_production, model.metabolites
        ) if np.isnan(solution) or solution < model.tolerance]
    return mets_not_produced
```

Comparison:
```python
if VERBOSITY:
    model.solver.interface.Configuration.verbosity = model.solver.interface.Configuration(
        verbosity=3
    )

print(f"Number of metabolites in the model: {len(model.metabolites)}")
start = time()
old = set([met.id for met in test_old(model)])
print(
    f"Identified metabolites by the old version: {len(old)} in {time() - start} s"
)
start = time()
new = set([met.id for met in find_metabolites_not_produced_with_open_bounds(model, PROCESSES)])
print(
    f"Identified metabolites by the new version: {len(new)} in {time() - start} s"
)
print(old - new)
print(new - old)

assert old == new
```
### Results
```
Number of metabolites in the model: 342
Identified metabolites by the old version: 22 in 2.826913356781006 s
Identified metabolites by the new version: 22 in 0.7438373565673828 s
set()
```

# 3. Test for consumption
The opposite test should also incorporate the new logic and the multiprocessing feature.
The corresponding tests:

```python
def consumed_old(model):
    """
    Return metabolites that cannot be consumed with open boundary reactions.
    When all metabolites can be secreted, it should be possible for each and
    every metabolite to be consumed in some form.
    Parameters
    ----------
    model : cobra.Model
        The metabolic model under investigation.
    Returns
    -------
    list
        Those metabolites that could not be consumed.
    """
    mets_not_consumed = list()
    open_exchanges(model)
    for met in model.metabolites:
        with model:
            exch = model.add_boundary(
                met, type="irrex", reaction_id="IRREX", lb=-1000, ub=0)
            solution = run_fba(model, exch.id, direction="min")
            if np.isnan(solution) or abs(solution) < model.tolerance:
                mets_not_consumed.append(met)
    return mets_not_consumed

def consumed_new(model, processes=None):
    return find_metabolites_not_produced_with_open_bounds(model, processes=processes, prod=False)
```

Comparison:
```python
if VERBOSITY:
    model.solver.interface.Configuration.verbosity = model.solver.interface.Configuration(
        verbosity=3
    )

print(f"Number of metabolites in the model: {len(model.metabolites)}")
start = time()
old = set(consumed_old(model))
print(
    f"Identified metabolites by the old version: {len(old)} in {time() - start} s"
)
old = set([met.id for met in old])
# old is 99 for this model in 12.536982536315918 s
start = time()
new = set(consumed_new(model, PROCESSES))
print(
    f"Identified metabolites by the new version: {len(new)} in {time() - start} s"
)
new = set([met.id for met in new])
print(old ^ new)

assert old == new
```

### Results
```
Number of metabolites in the model: 342
Identified metabolites by the old version: 20 in 2.8499839305877686 s
Identified metabolites by the new version: 20 in 0.8409318923950195 s
set()
```

src/memote/support/consistency.py
```diff
@@ -490,37 +492,122 @@ def find_disconnected(model):
...
+
+    Returns
+    -------
+    solution: float
+        the value of the solution of the LP problem, *NaN* if infeasible.
+    metabolite: cobra.Metabolite
+        metabolite passed as argument (to use map as a filter)
+
+    """
+    global _model
```
ChristianLieven: As far as I can tell `_init_worker` is always invoked before `_solve_metabolite_production` hence the declaration of these global variables is redundant (they are already global). As far as I recall you don't need to declare them as global to read their values.

carrascomj: Yes, you are right. Also, the changes doesn't affect speed of the multiprocessing feature, as noted in [this updated notebook](https://github.com/carrascomj/notebooks/blob/master/memote/test_consistency_prod_consumption_speed_up.ipynb).

ChristianLieven:
Oh before I merge could I get you to add an entry to the History.rst briefly describing the changes you made?

carrascomj:
OK, I'll add an entry as soon as possible!

PR #679, opened 2019-12-03
refactor: metabolite production/consumption

Midnighter:
I didn't get around to it in time, so here are some smaller refactorings on top of #678.

### DIFF
--- a/src/memote/support/consistency.py
+++ b/src/memote/support/consistency.py
@@ -594,7 +594,8 @@ def find_blocked_metabolites(model, coefficient, processes=None):
                 initargs=(model, sink.name, coefficient),
             )
             result_iter = pool.imap_unordered(
-                _solve_metabolite_exchange, met_identifiers, chunksize=chunk_size
+                _solve_metabolite_exchange, met_identifiers,
+                chunksize=chunk_size
             )
             pool.close()
             blocked = [met_id for solution, met_id in result_iter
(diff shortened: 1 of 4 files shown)


src/memote/support/consistency.py
```diff
@@ -20,15 +20,14 @@
...
 
 import numpy as np
-import multiprocessing
-from cobra import Reaction
+from cobra import Configuration, Reaction
 from cobra.exceptions import Infeasible
 from cobra.flux_analysis import flux_variability_analysis
-from cobra import Configuration
-from optlang.interface import OPTIMAL, INFEASIBLE
+from optlang.interface import INFEASIBLE, OPTIMAL
```
ChristianLieven: A change like this is super arcane to me. Does it matter in which order these two are imported?

src/memote/support/consistency.py
```diff
@@ -492,64 +493,124 @@ def find_disconnected(model):
...
     """
-    Add reaction to a `metabolite`'s contraints.
+    Solve for a metabolite's exchange flux.
 
-    The reaction and the model are passed as globals.
+    By adding the exchange variable to the metabolite constraint,
+    the solution tests whether the metabolic model produce or consume the
+    metabolite.
+
+    Notes
```
ChristianLieven: Nice touch!

src/memote/support/consistency.py
Midnighter: Honestly, I simply ran `isort` on the module that did this.

Issue #680, opened 2020-01-31
memote stuck with a *large* combined reconstruction


mfondi:
#### Problem description
We are trying to use memote on a multi-compartment, large reconstruction accounting for roughly 4500 reactions. We used the following code to launch memote on it:

docker run -v /home/matlab/Marco/memote/Models_For_Memote:/opt opencobra/memote memote report snapshot --filename "/opt/output_dir/report_OURmodel.html" /opt/OURmodel.xml

memote got stuck at this point (I killed the process after two days):

============================= test session starts ==============================
platform linux -- Python 3.6.9, pytest-5.3.1, py-1.8.0, pluggy-0.13.1
rootdir: /usr/local/lib/python3.6/site-packages/memote/suite/tests
collected 262 items

../../usr/local/lib/python3.6/site-packages/memote/suite/tests/test_annotation.py F [  0%]
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF.FFFFF.FFFFFFFFFFFFFFFFF         [ 24%]
../../usr/local/lib/python3.6/site-packages/memote/suite/tests/test_basic.py . [ 25%]
.....FF.....FFF.F.F.FF                                                   [ 33%]
../../usr/local/lib/python3.6/site-packages/memote/suite/tests/test_biomass.py . [ 33%]
FFFFFFFFFFFFFF..F...........FFFFFFFFFFFFFF.F
F.FFFF.F....FFFFFFFFFFFFFFFF [ 61%]
FFFFFFFFFFFFFFFFFFFFFFFFFF..............FFFFFFFFFFFFFF                   [ 82%]
../../usr/local/lib/python3.6/site-packages/memote/suite/tests/test_consistency.py F [ 82%]
ssssssssssssssss.FFFFF.

Do you have any suggestion on how to overcome this issue? Could it be that the reconstruction is too big for memote to analyse it? Is there an upper limit for the size of the reconstructions that memote can analyse?

I can provide more details if needed.

Thanks in advance for your help.
bets
m.

Midnighter:
Hi @mfondi, Thank you for trying out memote. There are a few tests that are notoriously slow for large models. I would recommend skipping them.

You don't show the traceback of interrupting above which would tell me exactly which test it is stuck on. I can guess that it is metabolite production or consumption.

[...]

PR #682, opened 2020-02-10
feat: accept solver timeout as parameter on CLI

carrascomj:
* [x] workaround #680 
`test_find_reactions_unbounded_flux_default_condition` in test_consistency can get stuck due to numerical instabilities while performing FVA.
* [x] Description of feature
Add a parameter on the CLI to allow the user to provide a solver timeout value.
```
memote --timeout-solver 3 run [...]
```
With this timeout, the test doesn't get stuck and continues. Whenever the solver reaches the timeout, it outputs an `UserWarning`. The problem is that timeout reactions won't be catch as Infeasible by the FVA implementation on cobrapy.

* [x] Decide to set a default value on timeout or temporarily set a value only for `test_find_reactions_unbounded_flux_default_condition` if none was provided.

* [x] tests added/passed

### DIFF
--- a/src/memote/suite/cli/config.py
+++ b/src/memote/suite/cli/config.py
@@ -51,6 +51,7 @@ class ConfigSectionSchema(object):
             type=click.Choice(["cplex", "glpk", "gurobi", "glpk_exact"]),
             default="glpk"
         )
+        timeout_solver = Param(type=int)
         experimental = Param(type=click.Path(exists=False, dir_okay=False),
                              default=None)
 
--- a/src/memote/suite/cli/config.py
+++ b/src/memote/suite/cli/config.py
@@ -51,7 +51,7 @@ class ConfigSectionSchema(object):
             type=click.Choice(["cplex", "glpk", "gurobi", "glpk_exact"]),
             default="glpk"
         )
-        timeout_solver = Param(type=int)
+        solver_timeou = Param(type=int, default=None)
         experimental = Param(type=click.Path(exists=False, dir_okay=False),
                              default=None)
 
--- a/src/memote/suite/cli/config.py
+++ b/src/memote/suite/cli/config.py
@@ -51,7 +51,7 @@ class ConfigSectionSchema(object):
             type=click.Choice(["cplex", "glpk", "gurobi", "glpk_exact"]),
             default="glpk"
         )
-        solver_timeou = Param(type=int, default=None)
+        solver_timeout = Param(type=int, default=None)
         experimental = Param(type=click.Path(exists=False, dir_okay=False),
                              default=None)
 
--- a/src/memote/suite/cli/reports.py
+++ b/src/memote/suite/cli/reports.py
@@ -69,7 +69,7 @@ def report():
               type=click.Choice(["cplex", "glpk", "gurobi", "glpk_exact"]),
               default="glpk", show_default=True,
               help="Set the solver to be used.")
-@click.option("--solver_timeout",
+@click.option("--solver-timeout",
               type=int, default=None,
               help="Timeout in seconds to set on the mathematical "
                    "optimization solver.")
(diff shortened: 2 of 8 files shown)


Midnighter:
This would be really neat to make an attribute on the `cobrapy.Configuration` class, we could then set it here through the command line option.

carrascomj:
On the cobrapy side, it would also be really helpful to slightly modify FVA to handle the timeout case. Right now, for every reaction, [the implementation only checks if the value is `None`](https://github.com/opencobra/cobrapy/blob/devel/cobra/flux_analysis/variability.py#L54) and that's not the case here, where the model.solver.status is set to `time_out` but the value of the solution is still a float.

Midnighter:
Probably it's better discussed in an issue on cobrapy but I don't fully see your point. In the line above that one

```python
value = _model.solver.objective.value
```

if the solver experienced a time-out, this should be a valid value, shouldn't it?

carrascomj:
Yes, it is a valid value but I am not sure if it should be considered a valid solution in the FVA loop or it should be discarded.

src/memote/suite/cli/runner.py
```diff
@@ -111,6 +111,9 @@ def cli():
               type=click.Choice(["cplex", "glpk", "gurobi", "glpk_exact"]),
               default="glpk", show_default=True,
               help="Set the solver to be used.")
+@click.option("--timeout-solver",
+              type=int,
+              help="Timeout to put in the LP solver.")
```
Midnighter:
```suggestion
              help="Timeout in seconds to set on the mathematical optimization solver.")
```

carrascomj:
I think the `test_model` test with a big model like you said would be problematic: the same test would potentially pass on Travis but fail on a local, faster, computer.

I believe that a better test would be to use a model that is known to get stuck on FVA. Another way is to set the timeout to 0 so another model wouldn't be needed.

src/memote/suite/cli/config.py
```diff
@@ -51,7 +51,7 @@ class Memote(SectionSchema):
             type=click.Choice(["cplex", "glpk", "gurobi", "glpk_exact"]),
             default="glpk"
         )
-        timeout_solver = Param(type=int)
+        solver_timeou = Param(type=int, default=None)
```
Midnighter:
```suggestion
        solver_timeout = Param(type=int, default=None)
```

src/memote/suite/cli/reports.py
```diff
@@ -69,6 +69,10 @@ def report():
               type=click.Choice(["cplex", "glpk", "gurobi", "glpk_exact"]),
               default="glpk", show_default=True,
               help="Set the solver to be used.")
+@click.option("--solver_timeout",
```
Midnighter:
```suggestion
@click.option("--solver-timeout",
```

src/memote/suite/cli/reports.py
```diff
@@ -204,6 +210,10 @@ def _test_diff(model_and_model_ver_tuple, pytest_args, skip,
 @click.option("--solver", type=click.Choice(["cplex", "glpk", "gurobi"]),
               default="glpk", show_default=True,
               help="Set the solver to be used.")
+@click.option("--solver_timeout",
```
Midnighter:
```suggestion
@click.option("--solver-timeout",
```

src/memote/suite/cli/runner.py
```diff
@@ -111,9 +111,10 @@ def cli():
               type=click.Choice(["cplex", "glpk", "gurobi", "glpk_exact"]),
               default="glpk", show_default=True,
               help="Set the solver to be used.")
-@click.option("--timeout-solver",
-              type=int,
-              help="Timeout to put in the LP solver.")
+@click.option("--solver_timeout",
```
Midnighter:
```suggestion
@click.option("--solver-timeout",
```

src/memote/suite/cli/runner.py
```diff
@@ -297,6 +298,10 @@ def _test_history(model, sbml_ver, solver, manager, commit, pytest_args, skip,
 @click.option("--solver", type=click.Choice(["cplex", "glpk", "gurobi"]),
               default="glpk", show_default=True,
               help="Set the solver to be used.")
+@click.option("--solver_timeout",
```
Midnighter:
```suggestion
@click.option("--solver-timeout",
```

carrascomj:
I have been trying some things with the test but in the end I've just mark it as xfailed. Maybe the only option is to try to work with a very large model like you said @Midnighter.

PR #684, opened 2020-04-06
fix: set threshold to model tolerance when solution is infeasible

carrascomj:
* [x] description of bug
The growth tests with experimental data do not work when, after applying the medium, the solver returns infeasible, propagating a `NaN` to the growth comparison that hampers the tests (reports 0 for all metabolites).
Now the user can provide an absolute value in the `experiments.yml` configuration as minimum growth that acts both on essentiality and growth experiments. If no value is provided, it defaults to a coefficient (set as 10% for now) of the growth under default constraints.
* [ ] tests added/passed
* [x] document min_growth

### DIFF
--- a/docs/experimental_data.rst
+++ b/docs/experimental_data.rst
@@ -49,11 +49,11 @@ The file simply states that files relating to media, essentiality, and growth
 experiments can be found at directories stated under ``path:`` relative to the
 location of the file.
 
-Minimum growth threshold
-------------------------
+Minimal growth rate
+-------------------
 Both growth and essentiality experiments require a threshold for the biomass
 function to determine if the model is growing. This value can be set via the
-``min_growth`` option in the ``experiments.yml`` file.
+``minimal_growth_rate`` option in the ``experiments.yml`` file.
 For instance:
 
 .. code-block:: yaml
@@ -65,9 +65,9 @@ For instance:
       path: "essentiality/"
     growth:
       path: "growth/"
-    min_growth: 0.05
+    minimal_growth_rate: 0.05
 
-By default, ``min_growth`` is set to 10% of the biomass function value
+By default, ``minimal_growth_rate`` is set to 10% of the biomass function value
 under the default constraints in the model.
 
 Media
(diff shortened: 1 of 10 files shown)


docs/experimental_data.rst
```diff
@@ -49,6 +49,27 @@ The file simply states that files relating to media, essentiality, and growth
 experiments can be found at directories stated under ``path:`` relative to the
 location of the file.
 
+Minimum growth threshold
+------------------------
+Both growth and essentiality experiments require a threshold for the biomass
+function to determine if the model is growing. This value can be set via the
+``min_growth`` option in the ``experiments.yml`` file.
```
Midnighter, line 56:
```suggestion
Both growth and essentiality experiments require a minimal biomass
function value to determine if the model predicts viability. This value can be set via the
``minimal_growth_rate`` option in the ``experiments.yml`` file.
```

src/memote/experimental/config.py
```diff
@@ -176,3 +183,30 @@ def get_path(self, obj, default):
...
+    def get_min_growth(self, model, threshold=0.1):
+        """Calculate min growth default value or return input value.
+
+        This value is used to determine if a model is capable of growth under
+        certain experimental conditions.
+
+        Parameters
+        ----------
+        model : cobra.Model
+        threshold : float
```
Midnighter, line 196:
```suggestion
        threshold : float, optional
```

src/memote/experimental/config.py
```diff
@@ -176,3 +183,30 @@ def get_path(self, obj, default):
...
+            If no input is provided by the user the default value is set to a
+            coefficient `threshold` times the growth under default constraints.
+            Default: 0.1.
```
Midnighter, line 199:
```suggestion
            coefficient `threshold` times the growth under default constraints (default 0.1).
```

src/memote/experimental/config.py
```diff
@@ -176,3 +183,30 @@ def get_path(self, obj, default):
...
+
+        """
+        min_growth = self.config.get("min_growth")
+        if min_growth is None:
+            min_growth = model.slim_optimize() * threshold
+            if isnan(min_growth):
+                LOGGER.debug(
```
Midnighter, line 206: By default pytest will swallow this message but it is a significant error that I care about as a user of memote.

```suggestion
                LOGGER.error(
```

PR #688, opened 2020-06-11
fix: detect minimal uncoservable sets

### LINKED ISSUE #378: Fix the stoichiometric consistency tests

carrascomj:
* [x] fix #378 (issue number)
The detection of minimal uncoservable sets was not complying with experimental results.
**There was a constraint lacking in the MILP problem (y>=0)**. Moreover, the nullspace computation was verified with the sympy interface, which, by default, normalizes the nullspace yielding the same results as in the toy examples in [Gevorgyan et al., 2008](https://academic.oup.com/bioinformatics/article/24/19/2245/248822). This behavior is due to the definition of "zero". It yields the same minimal unconservable sets as the numpy implementation so it was left as it was.
* [x] tests added/passed

### DIFF
--- a/src/memote/suite/tests/test_consistency.py
+++ b/src/memote/suite/tests/test_consistency.py
@@ -27,7 +27,7 @@ from memote.utils import annotate, truncate, get_ids, wrapper
 import memote.support.consistency_helpers as con_helpers
 
 
-@annotate(title="Stoichiometric Consistency", format_type="count")
+@annotate(title="Stoichiometric Consistency", format_type="percent")
 def test_stoichiometric_consistency(model):
     """
     Expect that the stoichiometry is consistent.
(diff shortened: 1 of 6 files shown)


carrascomj:
[This notebook](https://github.com/carrascomj/doc_lab/blob/master/notebooks/memote/toy2_consistency.ipynb) shows that sympy obtains the same results by default as in the Figure 1 of [Gevorgyan et al., 2008](https://academic.oup.com/bioinformatics/article/24/19/2245/248822).

src/memote/suite/tests/test_consistency.py
```diff
@@ -64,6 +64,44 @@ def test_stoichiometric_consistency(model):
     assert is_consistent, ann["message"]
 
 
+@annotate(title="Inconsistent Minimal Sets", format_type="count")
```
Midnighter, line 67: We decided to merge this test case with the above. For that, change the `format_type` to `"percent"`. This should display the metric in the report (please check that it displays correctly).

The test data should then be changed as follows:
```python
ann["data"] = {
    "unconserved_metabolites": [...],
    "inconsistent_minimal_stoichiometries": [...]
}
ann["metric"] = len(ann["data"]["unconserved_metabolites"]) / len(model.metabolites)
```

src/memote/support/consistency.py
```diff
@@ -219,23 +217,28 @@ def find_inconsistent_min_stoichiometry(model, atol=1e-13):
     metabolites = sorted(internal_mets, key=get_id)
     stoich, met_index, rxn_index = con_helpers.stoichiometry_matrix(
         metabolites, reactions)
-    left_ns = con_helpers.nullspace(stoich.T)
-    # deal with numerical instabilities
-    left_ns[np.abs(left_ns) < atol] = 0.0
-    LOGGER.info("nullspace has dimension %d", left_ns.shape[1])
+    left_ns = con_helpers.nullspace(stoich.T, atol)
+    if len(left_ns.shape) > 1:
```
Midnighter, line 221: We decided that if the nullspace is empty, we can return early with a list of every unconserved metabolite in a singleton set.

BenjaSanchez:
@carrascomj @Midnighter did you observe any effects on performance from these changes? I just upgraded my memote from 0.10.2 to 0.11.0 and now it seems to get permanently stuck in `test_stoichiometric_consistency` when running the test suite on [yeastGEM](https://github.com/SysBioChalmers/yeast-GEM/blob/devel/ModelFiles/xml/yeastGEM.xml):

`
memote report snapshot yeastGEM.xml --solver-timeout 30
`

### Context

<details>

System Information
==================
OS         Windows
OS-release      10
Python       3.7.7

Package Versions
================
Jinja2                           2.11.2
click                             7.1.1
click-configfile                  0.2.3
click-log                         0.3.2
cobra                            0.18.1
cookiecutter                      1.7.2
depinfo                           1.5.3
equilibrator-api                 0.1.26
future                           0.18.2
gitpython                         3.1.1
goodtables                       2.4.12
importlib-resources               1.4.0
lxml                              4.5.0
memote                           0.11.0
numpydoc                          0.9.2
pandas                            1.0.3
pip                              20.1.1
pylru                             1.2.0
pytest                            5.4.1
requests                         2.23.0
ruamel.yaml                     0.16.10
setuptools          47.3.1.post20200622
six                              1.14.0
sqlalchemy                       1.3.16
sympy                             1.5.1
travis-encrypt                    1.1.2
wheel                            0.34.2
</details>

carrascomj:
@BenjaSanchez Yes, `test_stoichiometric_consistency` now runs a MILP problem for every unconserved metabolite, so it could be considerably slower for big reconstructions. The `solver-timeout` parameter is applied to every MILP problem (so it _should_ finish); I'll check if the solver configuration is being correctly propagated for that case.

BenjaSanchez:
@carrascomj @Midnighter  I did some more testing, and after installing `symengine`,  `test_stoichiometric_consistency` takes ~45 seconds (without it it's over an hour), making the whole test suite in yeast-GEM take ~12 minutes. So feel free to disregard my comment :) but maybe symengine could be a recommended dependency of memote? It's seems to significantly help with several tests.

PR #703, opened 2020-10-01
refactor: split stoichiometry consistency in 3 tests

### LINKED ISSUE #701: Put a cap on the number of minimal inconsistent stoichiometries
### Checklist

- [x] There are [no similar issues or pull requests](https://github.com/opencobra/memote/issues) for this yet.

### Is your feature related to a problem? Please describe it.
Related to #698 #697 #680.

The MILP problems performed by [`find_inconsistent_min_stoichiometry`](https://github.com/opencobra/memote/blob/develop/src/memote/support/consistency.py#L189) are producing some issues regarding timeouts and worker errors.

## Describe the solution you would like.

I would suggest for putting a cap on the number of unconserved metabolites that memotes processes ([this loop](https://github.com/opencobra/memote/blob/develop/src/memote/support/consistency.py#L242)) to compute the minimal inconsistent stoichiometries and report it accordingly. This fix would make it easier (or possible) to use it for large networks locally or in the server.

## Describe alternatives you considered

As mentioned in https://github.com/opencobra/memote/pull/688#issuecomment-652974788 It seems like using CPLEX and symengine could be recommended but an error was reported anyways.

carrascomj:
* [ ] fix #701 
Split the stoichiometric consistency problematic tests into three and put a cap on the number of inconsistent stoichiometries to be found so that it does not take forever on large reconstructions.
* [x] tests added/passed

### DIFF
--- a/src/memote/suite/tests/test_consistency.py
+++ b/src/memote/suite/tests/test_consistency.py
@@ -27,7 +27,7 @@ import memote.support.consistency_helpers as con_helpers
 from memote.utils import annotate, get_ids, truncate, wrapper
 
 
-@annotate(title="Stoichiometric Consistency", format_type="percent")
+@annotate(title="Stoichiometric Consistency", format_type="raw")
 def test_stoichiometric_consistency(model):
     """
     Expect that the stoichiometry is consistent.
@@ -54,10 +54,61 @@ def test_stoichiometric_consistency(model):
     """
     ann = test_stoichiometric_consistency.annotation
     is_consistent = consistency.check_stoichiometric_consistency(model)
+    ann["data"] = is_consistent
+    ann["metric"] = 1.0 - float(is_consistent)
+    ann["message"] = wrapper.fill(
+        """This model is stoichiomerically {}""".format(
+            "consistent" if is_consistent else "inconsistent"
+        )
+    )
+    assert is_consistent, ann["message"]
+
+
+@annotate(title="Uncoserved Metabolites", format_type="percent")
+def test_unconserved_metabolites(model):
+    """
+    Report number of unconserved metabolites (if model incosistent).
+
+    List of unconserved metabolites is computed using the algorithm described
+    in section 3.2 of
+    "Detection of Stoichiometric Inconsistencies in Biomolecular Models."
+    Bioinformatics 24, no. 19 (2008): 2245.
+    doi: 10.1093/bioinformatics/btn425.
+    """
+    ann = test_stoichiometric_consistency.annotation
+    is_consistent = consistency.check_stoichiometric_consistency(model)
     ann["data"] = {
         "unconserved_metabolites": []
         if is_consistent
         else get_ids(consistency.find_unconserved_metabolites(model)),
+    }
+    ann["metric"] = len(ann["data"]["unconserved_metabolites"]) / len(model.metabolites)
+    ann["message"] = wrapper.fill(
+        """This model contains {} ({:.2%}) unconserved
+        metabolites: {}""".format(
+            len(ann["data"]["unconserved_metabolites"]),
+            ann["metric"],
+            truncate(ann["data"]["unconserved_metabolites"]),
+        )
+    )
+    # equivalent to is_consistent
+    assert ann["metric"] == 0, ann["message"]
+
+
+@annotate(title="Minimal Inconsistent Stoichiometries", format_type="count")
+def test_inconsistent_min_stoichiometry(model):
+    """
+    Report number of inconsistent min stoicihometries (if model incosistent).
+
+    Implementation:
+    Algorithm described in section 3.3 of
+    "Detection of Stoichiometric Inconsistencies in Biomolecular Models."
+    Bioinformatics 24, no. 19 (2008): 2245.
+    doi: 10.1093/bioinformatics/btn425.
+    """
+    ann = test_stoichiometric_consistency.annotation
+    is_consistent = consistency.check_stoichiometric_consistency(model)
+    ann["data"] = {
         "minimal_unconservable_sets": []
         if is_consistent
         else [
@@ -65,18 +116,14 @@ def test_stoichiometric_consistency(model):
             for mets in consistency.find_inconsistent_min_stoichiometry(model)
         ],
     }
-    ann["metric"] = len(ann["data"]["unconserved_metabolites"]) / len(model.metabolites)
+    ann["metric"] = len(ann["data"]["minimal_unconservable_sets"])
     ann["message"] = wrapper.fill(
-        """This model contains {} ({:.2%}) unconserved
-        metabolites: {}; and {} minimal unconservable sets: {}""".format(
-            len(ann["data"]["unconserved_metabolites"]),
-            ann["metric"],
-            truncate(ann["data"]["unconserved_metabolites"]),
+        """This model contains {} minimal unconservable sets: {}""".format(
             len(ann["data"]["minimal_unconservable_sets"]),
             truncate(ann["data"]["minimal_unconservable_sets"]),
         )
     )
-    assert is_consistent, ann["message"]
+    assert ann["metric"] == 0, ann["message"]
 
 
 @pytest.mark.parametrize("met", [x for x in consistency.ENERGY_COUPLES])
--- a/src/memote/suite/tests/test_consistency.py
+++ b/src/memote/suite/tests/test_consistency.py
@@ -27,7 +27,7 @@ import memote.support.consistency_helpers as con_helpers
 from memote.utils import annotate, get_ids, truncate, wrapper
 
 
-@annotate(title="Stoichiometric Consistency", format_type="raw")
+@annotate(title="Stoichiometric Consistency", format_type="percent")
 def test_stoichiometric_consistency(model):
     """
     Expect that the stoichiometry is consistent.
@@ -41,15 +41,11 @@ def test_stoichiometric_consistency(model):
     either produce mass from nothing or consume mass from the model.
 
     Implementation:
-    This test first uses an implementation of the algorithm presented in
+    This test uses an implementation of the algorithm presented in
     section 3.1 by Gevorgyan, A., M. G Poolman, and D. A Fell.
     "Detection of Stoichiometric Inconsistencies in Biomolecular Models."
     Bioinformatics 24, no. 19 (2008): 2245.
     doi: 10.1093/bioinformatics/btn425
-    Should the model be inconsistent, then the list of unconserved metabolites
-    is computed using the algorithm described in section 3.2 of the same
-    publication. In addition, the list of min unconservable sets is computed
-    using the algorithm described in section 3.3.
 
     """
     ann = test_stoichiometric_consistency.annotation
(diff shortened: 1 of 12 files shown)


carrascomj:
Need further investigation. The yeast-GEM model is still giving **Errored** in the report for Unconserved metabolites and  Minimal Inconsistent Errors, while the tests look good (they report the proper `AssertionError` with their corresponding messages after running `memote report snapshot`). It may be something with the configuration or the types of the metrics.

src/memote/suite/tests/test_consistency.py
```diff
@@ -51,32 +51,82 @@ def test_stoichiometric_consistency(model):
     publication. In addition, the list of min unconservable sets is computed
     using the algorithm described in section 3.3.
 
+    """
+    ann = test_stoichiometric_consistency.annotation
+    is_consistent = consistency.check_stoichiometric_consistency(model)
+    ann["data"] = is_consistent
+    ann["metric"] = 1.0 - float(is_consistent)
+    ann["message"] = wrapper.fill(
+        """This model is stoichiomerically {}""".format(
```
Midnighter, line 60:
```suggestion
        """The model's stoichiometry is {}""".format(
```

src/memote/suite/tests/test_consistency.py
```diff
@@ -51,32 +51,82 @@ def test_stoichiometric_consistency(model):
...
+            "consistent" if is_consistent else "inconsistent"
+        )
+    )
+    assert is_consistent, ann["message"]
+
+
+@annotate(title="Uncoserved Metabolites", format_type="percent")
```
Midnighter, line 67: As mentioned, I'd move this into the unscored part of the report. Thus a good format type would be `"count"`.

src/memote/suite/tests/test_consistency.py
```diff
@@ -51,32 +51,82 @@ def test_stoichiometric_consistency(model):
...
+def test_unconserved_metabolites(model):
+    """
+    Report number of unconserved metabolites (if model incosistent).
```
Midnighter, line 70:
```suggestion
    Report all unconserved metabolites.
```

src/memote/suite/tests/test_consistency.py
```diff
@@ -51,32 +51,82 @@ def test_stoichiometric_consistency(model):
...
+
+    List of unconserved metabolites is computed using the algorithm described
```
Midnighter, line 72:
```suggestion
    The list of unconserved metabolites is computed using the algorithm described
```

src/memote/suite/tests/test_consistency.py
```diff
@@ -51,32 +51,82 @@ def test_stoichiometric_consistency(model):
...
+    Bioinformatics 24, no. 19 (2008): 2245.
+    doi: 10.1093/bioinformatics/btn425.
     """
     ann = test_stoichiometric_consistency.annotation
     is_consistent = consistency.check_stoichiometric_consistency(model)
     ann["data"] = {
         "unconserved_metabolites": []
         if is_consistent
         else get_ids(consistency.find_unconserved_metabolites(model)),
+    }
```
Midnighter, line 84: Now that we have separate tests, this can go directly into data, I think.

```suggestion
    if is_consistent:
        ann["data"] = []
    else:
        ann["data"] = get_ids(consistency.find_unconserved_metabolites(model))
```

Midnighter: Please adjust the rest of the test accordingly.

src/memote/suite/tests/test_consistency.py
```diff
@@ -51,32 +51,82 @@ def test_stoichiometric_consistency(model):
...
+        )
+    )
+    # equivalent to is_consistent
+    assert ann["metric"] == 0, ann["message"]
+
+
+@annotate(title="Minimal Inconsistent Stoichiometries", format_type="count")
+def test_inconsistent_min_stoichiometry(model):
+    """
+    Report number of inconsistent min stoicihometries (if model incosistent).
```
Midnighter, line 101:
```suggestion
    Report inconsistent minimal net stoicihometries.
```

src/memote/suite/tests/test_consistency.py
```diff
@@ -51,32 +51,82 @@ def test_stoichiometric_consistency(model):
...
+    ann = test_stoichiometric_consistency.annotation
+    is_consistent = consistency.check_stoichiometric_consistency(model)
+    ann["data"] = {
         "minimal_unconservable_sets": []
         if is_consistent
         else [
             get_ids(mets)
             for mets in consistency.find_inconsistent_min_stoichiometry(model)
         ],
     }
```
Midnighter, line 121: Please put the outcome directly under the data key as in the previous test.

src/memote/suite/tests/test_consistency.py
```diff
@@ -51,32 +51,82 @@ def test_stoichiometric_consistency(model):
...
-    ann["metric"] = len(ann["data"]["unconserved_metabolites"]) / len(model.metabolites)
+    ann["metric"] = len(ann["data"]["minimal_unconservable_sets"])
```
Midnighter, line 122: Adjust this according to the change above.

src/memote/suite/tests/test_consistency.py
```diff
@@ -51,32 +51,82 @@ def test_stoichiometric_consistency(model):
...
+            ann["metric"],
+            truncate(ann["data"]["unconserved_metabolites"]),
+        )
+    )
+    # equivalent to is_consistent
+    assert ann["metric"] == 0, ann["message"]
+
+
+@annotate(title="Minimal Inconsistent Stoichiometries", format_type="count")
+def test_inconsistent_min_stoichiometry(model):
```
Midnighter, line 99: I just looked into the paper again and I think this test on its own actually doesn't deliver any insight. Only taken together with the elementary leakage modes described in section 3.4 do we find out more, i.e., the actual set of reactions contributing to the inconsistency.

> As an alternative solution for large models, we propose, instead of
calculating a complete set of elementary modes, calculating only a spanning
set that comprises the columns of a nullspace matrix of the system (N|y)v = 0;
this matrix can be obtained by Gauss–Jordan elimination. The inspection of
this matrix enables the detection of at least some, if not all of the input
errors causing the leakage of a given metabolite set (Fig. 2d). After the error
correction, the matrix can be recalculated to detect the remaining errors.

Would you be up for adding that? Probably best handled in a separate PR so that this can proceed.

src/memote/support/consistency.py
```diff
@@ -247,11 +251,15 @@ def find_inconsistent_min_stoichiometry(model, atol=1e-13):
             # singleton set!
             inc_minimal.add((met,))
             continue
+        if n_computed >= max_mets_computed:
```
Midnighter, line 254: Do you have a clear picture what is taking so long for large models? Is it because there are many unconserved metabolites or are the net stoichiometries large sets and many integer cuts (line 279) are applied?

I think this is also a candidate to be improved in a next PR. When we can report leakage modes, I guess, having a handful of them reported is enough to get started on improving the model. However, it would be best to try this out with an inconsistent model ourselves.

Midnighter (changes requested):
Cool, I hope combined with maybe a default timeout this will improve the situation a lot. I'd like to switch around the tests a bit so that not all of them are in the scored section.

src/memote/suite/tests/test_consistency.py
carrascomj: Maybe it is not that informative for the user without the modes. I can try to do that in another PR.

src/memote/support/consistency.py
carrascomj: It is hard to tell because both tests give an `Error` in the index. I tried the yeastGEM model (1904 uncoserved metabolites) with the cap, GLPK and without symengine and it was rather fast (the inconsistent minimal stoichiometry test is the one that takes most of the time among the three). I have removed the cap and will test it further. Do you have any idea why it gives the "Error"? The test since to retrieve a reasonable result so I am wondering if it has to do with the format.

carrascomj:
Sorry for the delay, I've answered to your comments in the review.

src/memote/suite/tests/test_consistency.py
Midnighter: Great, that'd be stellar!

src/memote/support/consistency.py
Midnighter: What do you mean exactly by

> both tests give an Error in the index

which index are you talking about?

carrascomj: Oh, sorry, I mean the report (I was looking at the _index.html_).

Midnighter:
What's your sense, do we need some more testing on different models? Should we include a default timeout to the command line interface?

src/memote/support/consistency.py
Midnighter: I see, I'll have to try out your branch.

carrascomj:
Yes, I think going with an opt-out 10 or 30 timeout is going to improve user experience. It is very unlikely to have an optimization of 10 seconds so we can just include in the documentation how to disable or change it for special cases. 
Anyways, I would like to have the thing with the error sorted out before merging this. To reproduce it:
```shell
curl -L -O https://raw.githubusercontent.com/SysBioChalmers/yeast-GEM/devel/ModelFiles/xml/yeastGEM.xml
memote -v DEBUG report snapshot --exclusive "test_unconserved_metabolites" yeastGEM.xml
```

src/memote/suite/tests/test_consistency.py
```diff
@@ -41,42 +41,81 @@ def test_stoichiometric_consistency(model):
...
+    """
+    Report number all unconserved metabolites.
+
+    The list of unconserved metabolites is computed using the algorithm described
+    in section 3.2 of
+    "Detection of Stoichiometric Inconsistencies in Biomolecular Models."
+    Bioinformatics 24, no. 19 (2008): 2245.
+    doi: 10.1093/bioinformatics/btn425.
+    """
+    ann = test_stoichiometric_consistency.annotation
```
Midnighter, line 74: Results are stored on the wrong test annotation.

```suggestion
    ann = test_unconserved_metabolites.annotation
```

src/memote/suite/tests/test_consistency.py
```diff
@@ -41,42 +41,81 @@ def test_stoichiometric_consistency(model):
...
+    Only 10 unconserved metabolites are reported and considered to
+    avoid computing for too long.
+
+    Implementation:
+    Algorithm described in section 3.3 of
+    "Detection of Stoichiometric Inconsistencies in Biomolecular Models."
+    Bioinformatics 24, no. 19 (2008): 2245.
+    doi: 10.1093/bioinformatics/btn425.
+    """
+    ann = test_stoichiometric_consistency.annotation
```
Midnighter, line 103: There is an error in the report because the data is stored on the wrong object and thus is missing here.

```suggestion
    ann = test_inconsistent_min_stoichiometry.annotation
```

src/memote/suite/tests/test_consistency.py
```diff
@@ -41,42 +41,81 @@ def test_stoichiometric_consistency(model):
...
+    ann["message"] = wrapper.fill(
+        """This model contains {} unconserved metabolites: {}""".format(
+            ann["metric"],
+            truncate(ann["data"]),
+        )
+    )
+    assert ann["metric"] == 0, ann["message"]
+
+
+@annotate(title="Minimal Inconsistent Stoichiometries", format_type="count")
```
Midnighter, line 89:
```suggestion
@annotate(title="Minimal Inconsistent Net Stoichiometries", format_type="count")
```

Midnighter:
I'm currently running both `test_unconserved_metabolites` and `test_inconsistent_min_stoichiometry` for the yeastGEM model and I think, we need to limit the number of investigated metabolites in `test_inconsistent_min_stoichiometry`. Most results look like singletons and (with CPLEX) the MIPs are solved very quickly (less than a second) but due to the sheer number it still takes a long time.

Midnighter:
I'm currently running both `test_unconserved_metabolites` and `test_inconsistent_min_stoichiometry` for the yeastGEM model and I think, we need to limit the number of investigated metabolites in `test_inconsistent_min_stoichiometry`. Most results look like singletons and (with CPLEX) the MIPs are solved very quickly (less than a second) but due to the sheer number it still takes a long time.

src/memote/suite/tests/test_consistency.py
carrascomj: Thank you, that solves a very annoying refactoring typo for me.

carrascomj:
I experience the same results with the yeast GEM. Putting a cap it took some time but it was quite less than test for open bounds so I think it was bearable. Shall we include it here or do I open another PR after this one?

Midnighter:
Please re-introduce the cap here. I also saw that many of the unconserved metabolites are external ones. This means, we probably fail to detect exchange reactions in the yeast model leading to the large number of unconserved metabolites. Something to look into but definitely not in this PR.

Midnighter (approved):
Great, thank you for staying on this issue.

PR #751, opened 2023-09-22
Update RegEx patterns for annotation IDs

WalkerKnapp:
* [x] Does not fix an open issue

This fix is related to issue #315, but it is a band-aid solution for the current state of database IDs, not an automatic solution as suggested in the issue.

The Regex patterns which are used to match against annotation IDs in `test_*_annotation_wrong_ids` tests have gone out of date since they were last updated. This PR replaces out-of-date ID patterns with the latest versions retrieved from [https://registry.identifiers.org/registry](https://registry.identifiers.org/registry).

I first noticed this problem when evaluating a model which uses a refseq id in the format `WP_*`, which is now a valid ID. While updating this pattern, I noticed that others have gone out of date as well and have updated them too.

* [x] tests added/passed

All previous tests in `tests/test_for_support/test_for_annotation.py` still apply and pass, and no additional functionality which would require tests has been introduced.


carrascomj (changes requested):
Thanks for the update! Would it be possible to add some identifiers to the tests that did not work before and now pass? Just to track the improvements (at one of [`met/rxn/gene_each_present`](https://github.com/opencobra/memote/blob/develop/tests/test_for_support/test_for_annotation.py#L75) fixture builders), it does not have to be exhaustive.

PR #756, opened 2023-11-01
Cycles met ids

### LINKED ISSUE #755: Generalize Metabolite IDs in test_detect_energy_generating_cycles
When I use MEMOTE on my models (generated in KBase), all of the "Erroneous Energy Generating Cycles" tests are skipped.
image

I determined that the problem was that my model's metabolites were not using the MetaNetX ID as the main ID, but most (>90%) of metabolites have a MetaNetID in their annotation field.
Because the MetaNetID is not use the metabolite ID, the tests were all being skipped because of this if statement.
https://github.com/opencobra/memote/blob/4c08f46849628507ac8c4de56b5d6ec745feed0b/src/memote/suite/tests/test_consistency.py#L156-L160

However, since the `detect_energy_generating_cycles` function retrieves the metabolites using the `find_met_in_model` function rather than just the ID, you can use the same function in the `test_detect_energy_generating_cycles` function to check that the metabolite is there before proceeding, such as:
```
main_comp = helpers.find_compartment_id_in_model(model, "c")
    try:
        helpers.find_met_in_model(model, met, main_comp)[0]
    except:
        pytest.skip(
            "This test has been skipped since metabolite {} could "
            "not be found in the model.".format(met)
        )
```

Doing that yielded the following results for my own model: 
image

I have made these changes on a branch of my own fork of the repo, and can open a pull request if this change would be helpful to others: https://github.com/hgscott/memote/tree/cycles_met_ids

hgscott:
* [x] fix #755
* [x] description of feature/fix- see #755 
* [ ] tests added/passed

### DIFF
--- a/src/memote/suite/tests/test_consistency.py
+++ b/src/memote/suite/tests/test_consistency.py
@@ -158,7 +158,9 @@ def test_detect_energy_generating_cycles(model, met):
     main_comp = helpers.find_compartment_id_in_model(model, "c")
     try:
         helpers.find_met_in_model(model, met, main_comp)[0]
-    except:
+    except ValueError:
+        pytest.skip("Metabolite {} is not in the MetaNetX shortlist.".format(met))
+    except (RuntimeError, IndexError):
         pytest.skip(
             "This test has been skipped since metabolite {} could "
             "not be found in the model.".format(met)
--- a/src/memote/suite/tests/test_consistency.py
+++ b/src/memote/suite/tests/test_consistency.py
@@ -154,12 +154,16 @@ def test_detect_energy_generating_cycles(model, met):
 
     """
     ann = test_detect_energy_generating_cycles.annotation
+    # Test if the metabolite is present in the model.
     main_comp = helpers.find_compartment_id_in_model(model, "c")
-    if helpers.find_met_in_model(model, met, main_comp)[0] is None:
+    try:
+        helpers.find_met_in_model(model, met, main_comp)[0]
+    except:
         pytest.skip(
             "This test has been skipped since metabolite {} could "
             "not be found in the model.".format(met)
         )
+    # If the metabolite is present, carry out the test.
     ann["data"][met] = consistency.detect_energy_generating_cycles(model, met)
     # Report the number of cycles scaled by the number of reactions.
     ann["metric"][met] = len(ann["data"][met]) / len(model.reactions)
--- a/src/memote/suite/tests/test_consistency.py
+++ b/src/memote/suite/tests/test_consistency.py
@@ -24,6 +24,7 @@ from cobra.flux_analysis import find_blocked_reactions
 
 import memote.support.consistency as consistency
 import memote.support.consistency_helpers as con_helpers
+import memote.support.helpers as helpers
 from memote.utils import annotate, get_ids, truncate, wrapper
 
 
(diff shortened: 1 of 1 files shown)


src/memote/suite/tests/test_consistency.py
```diff
@@ -153,11 +154,16 @@ def test_detect_energy_generating_cycles(model, met):
 
     """
     ann = test_detect_energy_generating_cycles.annotation
-    if met not in model.metabolites:
+    # Test if the metabolite is present in the model.
+    main_comp = helpers.find_compartment_id_in_model(model, "c")
```
Midnighter, line 158: There is probably a more general way to find the compartment, but "c" should work for most models. I must admit it's been a while for me to think about.

src/memote/suite/tests/test_consistency.py
```diff
@@ -153,11 +154,16 @@ def test_detect_energy_generating_cycles(model, met):
...
+    try:
+        helpers.find_met_in_model(model, met, main_comp)[0]
+    except:
```
Midnighter, line 161: It's a good idea to be more explicit about the expected exception.

```suggestion
    except ValueError:
        pytest.skip("Metabolite {} is not in the MetaNetX shortlist.".format(met))
    except (RuntimeError, IndexError):
```

Midnighter (changes requested):
Thank you for the changes! I'm curious to see all the problems this test will have once it is actually run on all the models where it was skipped before.

Please just make the changes for catching exceptions, the rest looks good!

Midnighter:
The tests fail because I need to make a fix to update the depinfo dependency... Not for you to worry about.

src/memote/suite/tests/test_consistency.py
hgscott: I agree. I did it this way, since it's what the `detect_energy_generating_cycles` function does, so that we know anything found here will be found in the function:
https://github.com/opencobra/memote/blob/4c08f46849628507ac8c4de56b5d6ec745feed0b/src/memote/support/consistency.py#L356

src/memote/suite/tests/test_consistency.py
hgscott: I know bare except statements are against the style guide, but specifying `ValueError` doesn't actually skip the tests where the metabolite isn't present.

Here is a screenshot from my report when L161 was `except ValueError:`:
image

And here is a screenshot for when L161 was just `except:`:
image

hgscott: Nevermind, you were right, I copied your code incorrectly!

hgscott:
> Thank you for the changes! I'm curious to see all the problems this test will have once it is actually run on all the models where it was skipped before.
> 
> Please just make the changes for catching exceptions, the rest looks good!

One thing to note that I found, because the `detect_energy_generating_cycles` functions returns all reactions where the flux is greater than 0.0, the list can be frighteningly long, but the vast majority of the reactions have a very low flux. So increasing the lower limit from 0.0 to something like 0.0001 decreases the number of reactions found by a lot (and probably gives a more helpful image of what is actually problematic).

Midnighter:
That's a good point. I suppose the flux should at least be greater than the tolerance of the solver.

Midnighter (approved):

Midnighter:
> One thing to note that I found, because the detect_energy_generating_cycles functions returns all reactions where the flux is greater than 0.0, the list can be frighteningly long, but the vast majority of the reactions have a very low flux. So increasing the lower limit from 0.0 to something like 0.0001 decreases the number of reactions found by a lot (and probably gives a more helpful image of what is actually problematic).

Now I merged your work without taking this into account. Would you be up for creating another PR to address this?

PR #768, opened 2024-10-21
Add option to output json format reports

LucoDevro:
* [x] Does not fix an open issue: it adds functionality

Json files are easier to parse for automated high-throughput applications. Since this output option was silently supported in the API, I exposed it via a Boolean flag.

* [x] tests added/passed

All tests passed (`make qa`).

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -5,6 +5,7 @@ Next Release
 ------------
 * Replace ``goodtables`` with ``pandera`` for data validation. This change is not
   100% backwards compatible, although most data tables should be unaffected.
+* Add option to output json report files.
 
 0.16.1 (2023-11-21)
 -------------------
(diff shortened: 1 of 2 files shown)


Midnighter:
Hi @LucoDevro,

Thank you for the contribution. I agree that this is very useful. It's been on my mind for years now, but I only get to work on memote in my spare time. I've held off on making this available so far, because I thought it would not be that useful without also offering a solution to load JSON files and creating a report from that.

Do you have a use just for this part already?

LucoDevro:
I'm setting up a high-throughput curation workflow and memote nicely puts together some tests. At some point, I needed to get the identifiers of the flagged reactions of course, and those are easier to get from a json than from an html.

Midnighter:
In that case, please also take a look at https://github.com/opencobra/jeweler which is a nextflow pipeline for running memote on many samples.

LucoDevro:
Thanks for the suggestion!

